PR #3, opened 2020-06-23
Start join html

EmmaWenzel:


src/main/webapp/join.html
```diff
@@ -0,0 +1,58 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>Plant Battles</title>
+    <link rel="stylesheet" href="style.css">
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css" integrity="sha384-9aIt2nRpC12Uk9gS9baDl411NQApFmC26EwAOH8WgZl5MYYxFfc+NcPb1dKGj7Sk" crossorigin="anonymous">
+    <script src="script.js"></script>
+    <script src="https://trefle.io/api/plants?q=rosemary&token=VU4zZmx5SnF3dmxVYTVsRlBTNGlqZz09"></script>
```
hannahsalameh, line 9: u can take this out if we aren't using trefle anymore

src/main/webapp/join.html
```diff
@@ -0,0 +1,58 @@
...
+              </form>
+              <br>
+              <p class="card-text">Want to join an already existing game? Submit the game ID here!</p>
+              <br>
+            </div>
+        </div>
+      </div>
+    <script src="https://code.jquery.com/jquery-3.5.1.slim.min.js" integrity="sha384-DfXdz2htPH0lsSSs5nCTpuj/zy4C+OGpamoFVy38MVBnE+IbbVYUew+OrCXaRkfj" crossorigin="anonymous"></script>
+    <script src="https://cdn.jsdelivr.net/npm/popper.js@1.16.0/dist/umd/popper.min.js" integrity="sha384-Q6E9RHvbIyZFJoft+2mJbHaEWldlvI9IOYy5n3zV9zzTtmI3UksdQRVvoxMfooAo" crossorigin="anonymous"></script>
+    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.5.0/js/bootstrap.min.js" integrity="sha384-OgVRvuATP1z7JjHLkuOU7Xw704+h835Lr+6QL9UvYjZE3Ipu6Tp75j7Bh/kR0JKI" crossorigin="anonymous"></script>
```
hannahsalameh, line 56: do u have to put the bootstrap scripts at the bottom of the page? I put mine in the header and I think it still works.

src/main/webapp/join.html
```diff
@@ -0,0 +1,58 @@
...
+                <a class="dropdown-item" href="join.html">Join a Game</a>
+                <a class="dropdown-item" href="start.html">Start a Game</a>
+                <div class="dropdown-divider"></div>
+                <a class="dropdown-item" href="#">Logout</a>
+                </div>
+            </li>
+          </ul>
+        </div>
+      </nav>
+      <br>
```
hannahsalameh, line 34: instead of adding all these br, u can just change the margin between the first div and the header. I think margin-top:50px would work. I can do this in the CSS too.

src/main/webapp/start.html
```diff
@@ -0,0 +1,60 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>Plant Battles</title>
+    <link rel="stylesheet" href="style.css">
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css" integrity="sha384-9aIt2nRpC12Uk9gS9baDl411NQApFmC26EwAOH8WgZl5MYYxFfc+NcPb1dKGj7Sk" crossorigin="anonymous">
+    <script src="script.js"></script>
+    <script src="https://trefle.io/api/plants?q=rosemary&token=VU4zZmx5SnF3dmxVYTVsRlBTNGlqZz09"></script>
```
hannahsalameh, line 9: same as other trefle comment

src/main/webapp/join.html
EmmaWenzel: oh, yep I missed that. Thanks!

src/main/webapp/join.html
EmmaWenzel: I put them in the bottom following the tutorial they had, but I think you're right they do work in the top and they look better there

src/main/webapp/join.html
EmmaWenzel: okay cool I was struggling with the margins before don't know if you can tell but it worked to change the margin like that (I did 200 pixels instead) and it made the background work a little better

src/main/webapp/join.html
devonwasson: https://stackoverflow.com/questions/30488298/where-in-the-html-file-should-jquery-and-bootstrap-be-placed

For reasoning on having something at the top vs the bottom.

src/main/webapp/join.html
```diff
@@ -0,0 +1,51 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>Plant Battles</title>
+    <link rel="stylesheet" href="style.css">
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css" integrity="sha384-9aIt2nRpC12Uk9gS9baDl411NQApFmC26EwAOH8WgZl5MYYxFfc+NcPb1dKGj7Sk" crossorigin="anonymous">
+    <script src="script.js"></script>
+    <script src="https://code.jquery.com/jquery-3.5.1.slim.min.js" integrity="sha384-DfXdz2htPH0lsSSs5nCTpuj/zy4C+OGpamoFVy38MVBnE+IbbVYUew+OrCXaRkfj" crossorigin="anonymous"></script>
```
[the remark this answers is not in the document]
EmmaWenzel: Should I delete the target files that I pushed in github?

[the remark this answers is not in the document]
devonwasson: yes, looks like you did it!

PR #7, opened 2020-06-30
Prepare blobstore

EmmaWenzel:


src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -0,0 +1,97 @@
...
+    Entity plantImageEntity = new Entity("plantImage");
+    plantImageEntity.setProperty("timestamp", System.currentTimeMillis());
+    // plantImageEntity.setProperty("username", username); 
+    plantImageEntity.setProperty("blobKey", blobKey);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.put(plantImageEntity);
+    
+    response.sendRedirect("/pictureUpload.html"); // TODO: send this somewhere else
+  }
```
colemanliyah, line 96: Can we put the do Get and do Post in separate files?

src/main/java/com/google/sps/servlets/ServeImageServlet.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.List;
+import java.util.Map;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
```
colemanliyah, line 34: I don't think you are using all these imports right? Could you get rid of some?

src/main/webapp/blobstoreFunctions.js
```diff
@@ -0,0 +1,54 @@
...
+
+/** Loads upload image URL and attaches it to the form */
+function fetchBlobstoreUrl() {
+  fetch('/blobstore-upload-url') 
+      .then((response) => {
+        return response.text();
+      })
+      .then((imageUploadUrl) => {
+        const messageForm = document.getElementById('plant-pic-form');
+        messageForm.action = imageUploadUrl;
```
colemanliyah, line 16: I think you could probably do document.getElementById('plant-pic-form).action = imageUploadUrl instead of creaiting a variable here, unless are you using this variable later for some reason?

src/main/webapp/blobstoreFunctions.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+  * Creates and adds a photo 
+  * pic is the URL of image to be shown in the post
+*/
+function createListElement(pic) {
+  let card = document.createElement("div");
+  card.className = "card m-1";
+
+  let body = document.createElement("div");
+  body.className = "card-body";
```
colemanliyah, line 30: Same comment as above to avoid making so many variables

src/main/webapp/blobstoreFunctions.js
```diff
@@ -0,0 +1,54 @@
...
+
+  let photo = document.createElement("img");
+  photo.style.width = "100%";
```
colemanliyah, line 33: could you maybe put this line in the css instead?

colemanliyah: line 33 i mean because I think if we use this line to make more images in the future we will want them to all have this style property

EmmaWenzel: should I just take this out for now, since we'll probably be using different size constraints when we add it to the quizzes? I could also make this the page where you can see the picture you currently have, but the whole page will probably need to be restyled. If that's the case I can add just this to the css for now

src/main/webapp/blobstoreFunctions.js
EmmaWenzel: These ones are each used twice, not sure if there's a way to consolidate creating the div, adding the class name and appending them together

src/main/webapp/blobstoreFunctions.js
EmmaWenzel: You're right, that works I'll change that

src/main/webapp/blobstoreFunctions.js
colemanliyah: If you don't need it then I would take it out, I wouldn't worry about creating another html page right now

src/main/webapp/blobstoreFunctions.js
colemanliyah: I what you mean, okay ignore this comment then lol

colemanliyah: I guess one thing I will just leave here is that maybe you could fix the spacing. I think there is a lot of spaces when some of the code could really be stacked on top of each other since it all goes together if that makes sense?

src/main/java/com/google/sps/servlets/ServeImageServlet.java
EmmaWenzel: I got rid of a chunk, all the other ones cause issues if I take them out. If something breaks in the future it might be because of the blobinfo imports but it seems to do fine without them here

src/main/webapp/blobstoreFunctions.js
EmmaWenzel: yep, I'll fix the spacing!

src/main/webapp/blobstoreFunctions.js
EmmaWenzel: sounds good, I took it out

src/main/java/com/google/sps/servlets/DataServlet.java
EmmaWenzel: yes, done!

src/main/java/com/google/sps/servlets/BlobstoreUploadUrlServlet.java
```diff
@@ -0,0 +1,40 @@
...
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * When the fetch() function requests the /blobstore-upload-url URL, the content of the response is
+ * the URL that allows a user to upload a file to Blobstore.
```
ofk-goog, line 27: This is an interesting idea, and I like how you documented it. But is the idea that the javascript code will retrieve a URL this way and then interact directly with Blobstore? It seems unsafe to permit the front-end to do this.

This is exactly the sort of decision that https://overthewire.org/wargames/natas will help you make.

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -0,0 +1,65 @@
...
+@WebServlet("/data")
+public class DataServlet extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    
+    // TODO: usernames
+    // String username = request.getParameter("username");
+
+    BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
```
ofk-goog, line 47: I'm guessing you can reduce latency noticeably by create this in the constructor instead of in doPost().

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -0,0 +1,65 @@
...
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.appengine.api.datastore.FetchOptions;
+
+/** Stores and retrieves posts from Datastore */
```
ofk-goog, line 37: Can you please be more specific about what a "post" is, or is that TBD?

src/main/java/com/google/sps/servlets/ServeImageServlet.java
ofk-goog: This confuses me. I'm not aware of how imports in java can have side effects. What error do you see if you don't import `MalformedURLException`?

src/main/java/com/google/sps/servlets/ServeImageServlet.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * When the user submits the form, Blobstore processes the file upload and then forwards the request
```
ofk-goog, line 37: What request is forwarded, from where? What exactly do you mean by forwarding a request? Can we find a way to make it easier to understand what's going on?

Actually, maybe there's a better place to put that information. In general, the documentation for Thing X should describe what Thing X does and how to use it. But it should avoid describing how Thing X is used by Thing Y, because it's likely that somebody will later change Thing Y without updating the documentation for Thing X.

src/main/java/com/google/sps/servlets/ServeImageServlet.java
EmmaWenzel: Just realized I got rid of extra imports in the wrong file. I went through all 4 files and checked all the imports, _now_ only necessary imports should be left

src/main/java/com/google/sps/servlets/DataServlet.java
EmmaWenzel: does constructor just mean above the doPost function but still in the class?

src/main/java/com/google/sps/servlets/DataServlet.java
EmmaWenzel: I changed it to "blob keys for plant images"

src/main/java/com/google/sps/servlets/DataServlet.java
ofk-goog: https://www.w3schools.com/java/java_constructors.asp

src/main/java/com/google/sps/servlets/ServeImageServlet.java
EmmaWenzel: How does
Serves the blob by passing the blob key to blobstore service; replaces the getServingUrl method
sound? My understanding of what this class does is still a little lacking, this was the part that was the work around for the errors in the original walkthrough. (From this documentation https://cloud.google.com/appengine/docs/standard/java/blobstore#Serving_a_blob)

src/main/java/com/google/sps/servlets/BlobstoreUploadUrlServlet.java
EmmaWenzel: I got the logic of this function from the other intern who had figured out how to work around the blobstore errors. From my understanding of the function, it's creating a url for the image that was uploaded and giving that back to the javascript (so changes in the javascript wouldn't create any points of contact between the javascript and the blobstore service). That said, I could be understanding it wrong and we haven't started playing the game yet. 
Is there a threat I'm missing?

ofk-goog: The doc says the response contains a "URL that allows a user to *upload* a file to blobstore". Whereas I think you're saying the returned URL can be used to *download* the file.

It looks to me like the former is true. In which case the user is able to upload whatever they want to blobstore.

PR #8, opened 2020-07-01
adapting style changes to this page - small pull request

colemanliyah:


src/main/webapp/gameBoard.html
```diff
@@ -1,91 +1,71 @@
...
+        </ul>
+        <ul class="nav navbar-nav ml-auto">
+        <li>UserName</li>
+        </ul>
     </div>
 </nav>
 
-<div class="view">
-    <body>
+<body onload="checkCookie()">
```
ofk-goog, line 34: What does checkCookie() do? I can't find where it's defined.

colemanliyah: Oops, I forgot to take this out. Just committed the new change to fix this

PR #9, opened 2020-07-06
add servlet to store blobkeys for images in datastore

EmmaWenzel:


src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.io.IOException;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
```
colemanliyah, line 27: are you using this import?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -0,0 +1,60 @@
...
+    BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+    ImmutableMap<String, List<BlobKey>> blobs = ImmutableMap.copyOf(blobstoreService.getUploads(request));
+    ImmutableList<BlobKey> blobKeys = ImmutableList.copyOf(blobs.get("image"));
+    String blobKey = blobKeys.get(0).getKeyString();
+ 
+    // TODO: might need a loop in here to check if that username already has a
+    // photo and replace it
+ 
+    Entity plantImageEntity = new Entity("plantImage");
+    plantImageEntity.setProperty("timestamp", System.currentTimeMillis());
```
colemanliyah, line 51: How come you are adding a time property to the plant images? Is it to help when cre

src/main/java/com/google/sps/servlets/DataServlet.java
EmmaWenzel: Yes, I get a compilation error without it

src/main/java/com/google/sps/servlets/DataServlet.java
EmmaWenzel: at the moment, I was using it to get the most recently uploaded image (to make sure that the pictures were uploading each time). I don't think we'll need this when we only have one picture per person, we'll just need a username attached instead of a timestamp. Now that I know it's working, should I take it out?

colemanliyah: Yeah, if you don't need it I would definitely say take it out

EmmaWenzel: got it! It's removed

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -0,0 +1,59 @@
...
+@WebServlet("/data")
+public class DataServlet extends HttpServlet {
+ 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    
+    // TODO: usernames
+    // String username = request.getParameter("username");
+ 
+    BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
```
ofk-goog, line 42: Can we reuse the same `BlobstoreService` instance instead of creating a new one every time?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -0,0 +1,59 @@
...
+    String blobKey = blobKeys.get(0).getKeyString();
+ 
+    // TODO: might need a loop in here to check if that username already has a
+    // photo and replace it
+ 
+    Entity plantImageEntity = new Entity("plantImage");
+    // plantImageEntity.setProperty("username", username); 
+    plantImageEntity.setProperty("blobKey", blobKey);
+ 
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
ofk-goog, line 54: Can we reuse a DatastoreService instance instead of making a new one?

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.io.IOException;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+ 
+/** Stores blob keys for plant images in Datastore */
```
ofk-goog, line 32: This is an upload handler, right? As described here: https://cloud.google.com/appengine/docs/standard/java/blobstore#3_implement_upload_handler

Please document that this servlet is a blobstore upload handler. (I only figured it out because you used BlobstoreService.getUploads().)

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -0,0 +1,59 @@
...
+@WebServlet("/data")
```
ofk-goog, line 33: I recommend choosing a more specific name, e.g. "/image_upload_blobstore_handler"

src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -0,0 +1,59 @@
...
+public class DataServlet extends HttpServlet {
+ 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
```
ofk-goog, line 38: Suggested TODO: Figure out how to limit this so it can only be used as a blobstore upload handler. It seems like bad news if an end-user can access this servlet directly.

src/main/java/com/google/sps/servlets/DataServlet.java
EmmaWenzel: I added a constructor to create this and the datastore service; I think that's what you meant but let me know if it's not.

src/main/java/com/google/sps/servlets/DataServlet.java
EmmaWenzel: I added the words upload handler, but I wanted to make sure there's nothing else it's doing that I'm missing. Is that first chunk of code just getting the most recently uploaded blobkey or is it manipulating the upload?

src/main/java/com/google/sps/servlets/DataServlet.java
EmmaWenzel: Got it, that's done. Should I change the name of the class/file too?

src/main/java/com/google/sps/servlets/DataServlet.java
EmmaWenzel: Sounds good, I added it.

src/main/java/com/google/sps/servlets/DataServlet.java
ofk-goog: I think so

src/main/java/com/google/sps/servlets/DataServlet.java
ofk-goog: Why are you asking me? You wrote the code. You are responsible for knowing what it does and communicating the important parts to other engineers.

ofk-goog: Sorry, that came out a little harsh. I don't mean that you have to understand everything without talking to other engineers. I just mean that the time to ask questions and figure this stuff out is before sending out a PR.

EmmaWenzel: Yeah, I kind of figured that while I was typing it. I thought I knew what it was doing (I assumed it was just grabbing the blobkey after it was placed by the uploadBlobstore servlet), but parts of the documentation you sent went over my head and I realized that might not be the case. In the future, I'll try to catch that before I send the pull request. Since this is already out, does my understanding make sense (does the new description make sense?)

ofk-goog: Your description is accurate and concise, and maybe a little terse. It might help to link to the documentation for blobstore upload handlers, because Liyah/Hannah/Devon might not know what that means otherwise.

EmmaWenzel: Sounds good, I'll do that.

src/main/java/com/google/sps/servlets/DataServlet.java
EmmaWenzel: Alright, it's renamed

PR #12, opened 2020-07-08
Created timing feature; eventually quizTiming servlet will be combined with game ID servlet

colemanliyah:


src/main/java/com/google/sps/servlets/getDailyQuiz.java
```diff
@@ -0,0 +1,61 @@
...
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery pq = datastore.prepare(query);
+
+        Object time = null;
+        for(Entity entity : pq.asIterable()) {
+            time = entity.getProperty("timestamp");
+        }
+
+        FetchOptions fetchOptions = FetchOptions.Builder.withLimit(1);        
+        List<Entity> quizEntities = pq.asQueryResultList(fetchOptions);
```
hannahsalameh, line 41: do u have to return it as a list if there is only 1 object you're fetching at a time?

src/main/java/com/google/sps/servlets/getDailyQuiz.java
```diff
@@ -0,0 +1,61 @@
...
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        
+        Query query = new Query("Quiz");
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery pq = datastore.prepare(query);
+
+        Object time = null;
+        for(Entity entity : pq.asIterable()) {
+            time = entity.getProperty("timestamp");
+        }
```
hannahsalameh, line 38: r u trying to get the time the quiz should release here? can u sort the datastore objects by time and get 1 value instead of iterating through all the times?

src/main/webapp/gameFunctions.js
```diff
@@ -0,0 +1,11 @@
+//This is the js file for gameBoard.html
+
+function getQuizStatus() {
+    fetch("/getQuizStatus").then(response => response.json()).then((tasks) => {
+        console.log(tasks);
```
hannahsalameh, line 5: i would take out console.logs

src/main/java/com/google/sps/servlets/getDailyQuiz.java
colemanliyah: I am not sure how to save it without errors if it is not in a list, do you know how this might be done? But also, I was thinking that sending a list to js makes everything easier to use/manipulate so keeping it a list might be good

src/main/java/com/google/sps/servlets/getDailyQuiz.java
colemanliyah: Actually, the idea is that there will be only one Quiz entity in the datastore so I am looping through this just to grab a specific property of the Quiz entity which here is just the time the quiz was made

src/main/java/com/google/sps/servlets/getDailyQuiz.java
hannahsalameh: One way to do it would be pq.asQueryResultList(fetchOptions)[1]; thats what i did in my code. if u think keeping a list is easier, thats fine too.

src/main/java/com/google/sps/servlets/getDailyQuiz.java
hannahsalameh: oooh ok so maybe do the thing i mentioned above and just get the one entity instead of the loop.

src/main/java/com/google/sps/servlets/getDailyQuiz.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.ArrayList;
+import java.util.List;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.QueryResultList;
+import java.text.DateFormat;
+import java.util.Date;
+import com.google.appengine.api.datastore.Key;
```
devonwasson, line 24: The following imports can likely safely be removed:

import com.google.gson.reflect.TypeToken;
import java.lang.reflect.Type;
import com.google.appengine.api.users.UserService;
import com.google.appengine.api.users.UserServiceFactory;
import java.util.ArrayList;
import java.util.List;
import com.google.appengine.api.datastore.QueryResultList;

As always though, remove these dependencies and rebuild to ensure they aren't actually needed.

src/main/java/com/google/sps/servlets/getDailyQuiz.java
```diff
@@ -0,0 +1,54 @@
...
+
+@WebServlet("/getQuizStatus")
+public class getDailyQuiz extends HttpServlet {
```
devonwasson, line 27: class name and Web Servlet path should be named the same.

src/main/java/com/google/sps/servlets/getDailyQuiz.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.servlet.http.HttpServletResponse;
+import java.util.ArrayList;
+import java.util.List;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.QueryResultList;
+import java.text.DateFormat;
+import java.util.Date;
+import com.google.appengine.api.datastore.Key;
+
+@WebServlet("/getQuizStatus")
```
devonwasson, line 26: File name and Web Servlet path should be similar, if not the same. In this case, having the file name be getDailyQuiz but the path be getQuizStatus is misleading. 

The only time these should be different is if we want to name the file getXYZ but we just call the path XYZ.

src/main/java/com/google/sps/servlets/getDailyQuiz.java
```diff
@@ -0,0 +1,54 @@
...
+public class getDailyQuiz extends HttpServlet {
+
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        
+        Query query = new Query("Quiz");
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery pq = datastore.prepare(query);
+
+        Entity quiz = pq.asList(FetchOptions.Builder.withLimit(1)).get(0);
+        Object time = quiz.getProperty("timestamp");
```
devonwasson, line 36: time as a variable name is a bit ambiguous and is generally iffy as many imports are called "time". Calling this variable something more specific (like "current_quiz_timestamp") will help with readability and ensure there are no weird bugs with naming conflicts in the future.

src/main/java/com/google/sps/servlets/getDailyQuiz.java
```diff
@@ -0,0 +1,54 @@
...
+@WebServlet("/getQuizStatus")
+public class getDailyQuiz extends HttpServlet {
+
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        
+        Query query = new Query("Quiz");
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        PreparedQuery pq = datastore.prepare(query);
+
+        Entity quiz = pq.asList(FetchOptions.Builder.withLimit(1)).get(0);
```
devonwasson, line 35: If I understand this correctly, there is only one quiz for the entire website? There is not one quiz generated per user? So everyone just gets the same quiz? And, if the day has rolled over, we generate a new quiz for users to take instead? That seems fine if this is the case, just wanted to make sure I'm understanding this correctly. If this is the case though, we could simplify this logic and make fewer database calls by storing the quiz as a variable in the class itself and do something like:

If class.quiz == null:
--check database for existing quiz
--if no existing quiz
----make new quiz
----store in class.quiz
----store in database
--else if existing quiz in database
----store existing quiz in class.quiz
else if class.quiz
--return class.quiz

src/main/java/com/google/sps/servlets/quizTiming.java
```diff
@@ -0,0 +1,32 @@
...
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.ArrayList;
+import java.util.List; 
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
```
devonwasson, line 15: The following imports could likely be safely removed:

import com.google.gson.Gson;
import java.util.ArrayList;
import java.util.List; 
import com.google.appengine.api.users.UserService;
import com.google.appengine.api.users.UserServiceFactory;

Be sure to rebuild after removing these though to ensure they actually are not needed.

src/main/java/com/google/sps/servlets/quizTiming.java
```diff
@@ -0,0 +1,32 @@
...
+
+@WebServlet("/quizes")
+public class quizTiming extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    Entity quiz = new Entity("Quiz");
+    quiz.setProperty("timestamp", System.currentTimeMillis());
+
```
devonwasson, line 25: So, if I understand correctly, this servlet creates a new quiz, right? It makes an assumption that there is no existing quiz. However, based on my comments in the getDailyQuiz.java file, we could likely combine this logic into a single servlet. 

Additionally, is this servlet being used? I don't see any new references to it in this PR.

src/main/webapp/gameFunctions.js
```diff
@@ -0,0 +1,9 @@
+//This is the js file for gameBoard.html
+
+function getQuizStatus() {
+    fetch("/getQuizStatus").then(response => response.json()).then((tasks) => {
+        if(tasks != "NoNewButton") {
```
devonwasson, line 5: It might be best to have this servlet return some actual json rather than just a hardcoded string. What if we change the string in the servlet to say "DontDisplayButton" for some reason? That would break this function. But if we have it return JSON, we can just check to see if the JSON is empty or not and go from there. So this if statement would be something like 

if (!tasks.isEmpty())

(I'm not sure of the exact syntax)

src/main/java/com/google/sps/servlets/getDailyQuiz.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Date;
+import com.google.appengine.api.datastore.Key;
+
+@WebServlet("/getQuizStatus")
+public class getDailyQuiz extends HttpServlet {
+
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        
+        Query query = new Query("Quiz");
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
devonwasson, line 32: Saving a single datastore object in the constructor of this class would speed things up rather than making a new instance every time this path is called.

src/main/java/com/google/sps/servlets/quizTiming.java
```diff
@@ -0,0 +1,32 @@
...
+@WebServlet("/quizes")
+public class quizTiming extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    Entity quiz = new Entity("Quiz");
+    quiz.setProperty("timestamp", System.currentTimeMillis());
+    
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
devonwasson, line 26: Saving a single datastore object in the constructor of this class would speed things up rather than making a new instance every time this path is called.

PR #14, opened 2020-07-08
Add servlet to create game entity for start game

EmmaWenzel:


src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -0,0 +1,62 @@
...
+ 
+      // create entity
+      Entity gameEntity = new Entity("Game");
+      gameEntity.setProperty("gameName", gameName);
+      gameEntity.setProperty("userNames", userNames);
+ 
+      // TODO: update user entity 
+      // Query query = new Query("User");
+      // PreparedQuery results = datastore.prepare(query);
+      // TODO: add game to user entity
```
colemanliyah, line 54: Do you think maybe we should add the user entity on a different servlet linked to the login page? Or maybe link the loginpage to this servlet but then change the name of the "/start-game" to something else?

src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -0,0 +1,65 @@
...
+ 
+  DatastoreService datastore;
+ 
+  public startGameServlet(){
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+ 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+      String gameName = request.getParameter("game-name");
```
devonwasson, line 41: Should probably check for existence of a valid game name here. What if someone sends over a game name that already exists (which might not be an issue globally, but would be an issue for a specific user... I probably dont want to have two active games with the same name)? or an empty game name? we should be able to detect these issues and then act accordingly. 

Additionally, this is a great candidate for test code. This could be its own method and we can have some nice test cases written around it.

EmmaWenzel: Okay, so that function is written, and I have a call to it commented out in this file, because it won't work until we incorporate user entities.

src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -38,7 +39,20 @@ public startGameServlet(){
...
       String gameName = request.getParameter("game-name");
+      String quizQuestion = "hardCodedTemp";
+      long quizTime = 0;
+
+      // TODO: check if game name is valid (needs user entity and GameUtils merged)
+      // Query query = new Query("User");
+      // PreparedQuery results = datastore.prepare(query);
+      // List<Entity> resultsList = results.asList();
+      // GameUtils gameUtils = new GameUtils();
+      // if(!gameUtils.isValidGameName(gameName, resultsList, givenUser)){
```
devonwasson, line 52: is it possible to query for a specific user name rather than return the entire list of user entities?

EmmaWenzel: I looked into it a little, it looks like you can use a filter like this:      
Filter getCorrectUser = new FilterPredicate("userID", FilterOperator.EQUAL, givenUser);
Query query = new Query("Game").setFilter(getCorrectUser);
PreparedQuery results = datastore.prepare(query);
I tried it out with my game entity and the game name property, and it returns the correct entity. I can't test if it will work with the user entity, but I'm assuming it will return the correct entity. I added that to the commented section

src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -45,11 +45,12 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       long quizTime = 0;
 
       // TODO: check if game name is valid (needs user entity and GameUtils merged)
-      // Query query = new Query("User");
+      // Filter getCorrectUser = new FilterPredicate("userID", FilterOperator.EQUAL, givenUser);
+      // Query query = new Query("Game").setFilter(getCorrectUser);
       // PreparedQuery results = datastore.prepare(query);
       // List<Entity> resultsList = results.asList();
```
devonwasson: This is really good. We could make it even better by refactoring it out into a function in a utility class (UserUtility?). that way, other people can use this bit of code (because I'm sure it'll be repeated again elsewhere. 

We can then also test it by mocking out the datastore and ensuring things work as expected. Some good test cases would be where the username doesn't exist, where the username exists, and where the given username is blank.

PR #15, opened 2020-07-08
Create User Entity

hannahsalameh:


src/main/java/com/google/sps/servlets/popUserEntity.java
```diff
@@ -0,0 +1,80 @@
...
+
+@WebServlet("/user")
+public class popUserEntity extends HttpServlet {
+DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+      String username = request.getParameter("name");
+      String userID = request.getParameter("id");
+      ArrayList<GameScores> userScoresArrayList = new ArrayList<GameScores>(); //this gets populated when the user joins a game.
```
colemanliyah, line 43: I think we might be able to take this line out. From Emma's servlet it seems that when a user starts a game it is in that servlet that the game ID will be added to the user entity

src/main/java/com/google/sps/servlets/popUserEntity.java
```diff
@@ -0,0 +1,80 @@
...
+
+    //check if user entity is already in system
+     Filter getCorrectUser = new FilterPredicate("userID", FilterOperator.EQUAL, userID);
+     Query query = new Query("user");
+     query.setFilter(getCorrectUser);
+
```
colemanliyah, line 49: Just delete one line here because of the extra space

src/main/java/com/google/sps/servlets/popUserEntity.java
```diff
@@ -0,0 +1,80 @@
...
+     PreparedQuery results = datastore.prepare(query);
+     
+     //this should always work, as long as doPost is always called upon login
+     try{
+         Entity userEntity = results.asList(FetchOptions.Builder.withLimit(1)).get(0);
+     } catch (Exception e){
+         System.err.println("user entity uninitialized. cannot return information.");
+     }
+  }
```
colemanliyah, line 79: could you put the doGet() into a separate java file? I think it will be cleaner and consistent with the rest of the code so far if we separate the doGet() and doPost() methods

src/main/java/com/google/sps/servlets/updateUser.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.gson.Gson;
+
+@WebServlet("/update-user")
+public class updateUser extends HttpServlet {
```
colemanliyah, line 30: what page will this servlet be connected to? What is the purpose of this servlet going to be?

hannahsalameh: oh whoops this page isnt necessary

src/main/webapp/loginScript.js
```diff
@@ -1,10 +1,21 @@
...
 }
 
+//create/update user entity
+function updateUserEntity(name, id){
+    var xhttp = new XMLHttpRequest();
+    xhttp.open("POST", "/user", false);
+    xhttp.setRequestHeader("Content-type", "application/x-www-form-urlencoded");
+    xhttp.send("name=" +name+ "&id=" +id);
+}
+
```
colemanliyah, line 18: are you sure we want to update the user's information in js? Would it be easier to do in a servlet and safer too? Also, how come we would need to update the user's information here? I thought this file just checks if the user is logged in or not

src/main/java/com/google/sps/servlets/popUserEntity.java
hannahsalameh: This line is just to initalize the arraylist for emma's servlet. I dont think u can add properties to the users list, so this just makes an empty list so emmas can put in the game id data

src/main/webapp/loginScript.js
hannahsalameh: i just took out the ID, but the only thing this is updating otherwise is the username in the corner which isn't so bad if the client manipulates it. also I realize the function is named bad, but really what it does is it creates a user entity for everyone logged in if they do not already have one. I changed the name of the function to reflect that a little more

hannahsalameh: its always checking the ID not the name, so having the name be fiddled w shouldnt change much

src/main/java/com/google/sps/servlets/popUserEntity.java
colemanliyah: okay do you think you could just change the ArrayList to be a Hashmap or some kind of dictionary where the key is the game id and then the individual scores will be the values?

src/main/webapp/loginScript.js
```diff
@@ -1,10 +1,20 @@
 function onSignIn(googleUser) {
   var profile = googleUser.getBasicProfile();
-  localStorage.setItem("username",profile.getName()); //sends player's names into local storage
-  localStorage.setItem("ID", profile.getId()); //sends player's id into local storage
+  var name = profile.getName();
+  var id = profile.getId();
```
ofk-goog, line 4: We're using this ID as a key in Datastore. But are you sure you get the same ID every time you log in with your Google account? If it is the same every time, then it's not safe to use it as a login cookie the way we have been doing.

src/main/java/com/google/sps/servlets/GameScores.java
```diff
@@ -0,0 +1,21 @@
+
+package com.google.sps.servlets;
```
ofk-goog, line 2: This should go in a separate package because it's not a Servlet.

src/main/java/com/google/sps/servlets/PopulateUserEntity.java
```diff
@@ -32,15 +32,26 @@
 import com.google.gson.Gson;
 
 @WebServlet("/user")
-public class popUserEntity extends HttpServlet {
+public class PopulateUserEntity extends HttpServlet {
 DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
+     /**
+    * Populates the datastore with "user" entities. 
+    * Every time someone logs in, this doPost is called
```
ofk-goog, line 40: "Every time someone logs in" -- You may decide to change this in the future, so I recommend not writing it down here, or else this comment will go out of date.

In general, I try to avoid describing the caller's behavior when documenting the callee. Sometimes it's useful to bend the rule by documenting *hypothetical* or *suggested* caller behavior. E.g. "For example, you could safely call this whenever a user logs in." Or *required* caller behavior, e.g. "This must be called at least once for every new user."

src/main/java/com/google/sps/servlets/PopulateUserEntity.java
```diff
@@ -32,15 +32,26 @@
 import com.google.gson.Gson;
 
 @WebServlet("/user")
```
ofk-goog, line 34: Please use a more specific path than "/user" -- I expect you'll have more than one user-related servlet.

PR #22, opened 2020-07-14
Add create game entity helper function

EmmaWenzel:
Create game entity helper function adds gameName, userIds, quizQuestion and timestamp to a new game and stores the game in datastore. The testing covers giving the function a null datastore instance or an empty string.

### DIFF
--- a/src/main/java/com/google/sps/utils/GameUtils.java
+++ b/src/main/java/com/google/sps/utils/GameUtils.java
@@ -20,8 +20,11 @@ import java.util.ArrayList;
 import com.google.appengine.api.datastore.EntityNotFoundException;
 import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.datastore.Key;
+import java.util.logging.Logger;
 
 public final class GameUtils {
+
+  private static final Logger log = Logger.getLogger(GameUtils.class.getName());
 
   /**
   * Checks to see whether the user already has a game by the given name, and returns false if so
@@ -53,7 +56,7 @@ public final class GameUtils {
         return null;
     }
     if(datastore == null){
-        System.out.println("Error: Null Datastore");
+        log.severe("null datastore");
         return null;
     }
     
@@ -80,7 +83,7 @@ public final class GameUtils {
       entity.setProperty("gameId", key);
       datastore.put(entity);
     }catch(EntityNotFoundException e){
-      System.out.println("Error: Entity Not Found Exception");
+      log.severe("EntityNotFoundException");
       return null;
     }
  
--- a/src/main/java/com/google/sps/utils/GameUtils.java
+++ b/src/main/java/com/google/sps/utils/GameUtils.java
@@ -71,12 +71,13 @@ public final class GameUtils {
     Entity entity = null;
     try{
       entity = datastore.get(gameKey);
+
+      String key = KeyFactory.keyToString(gameKey);
+      entity.setProperty("gameId", key);
+      datastore.put(entity);
     }catch(EntityNotFoundException e){
       return false;
     }
-    String key = KeyFactory.keyToString((entity.getKey()));
-    entity.setProperty("gameId", key);
-    datastore.put(entity);
  
     return true;
   }
--- a/src/main/java/com/google/sps/utils/GameUtils.java
+++ b/src/main/java/com/google/sps/utils/GameUtils.java
@@ -17,7 +17,6 @@ package com.google.sps.utils;
 import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.DatastoreService;
 import java.util.ArrayList;
-import java.util.List;
 
 
 public final class GameUtils {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -40,4 +43,41 @@ public boolean IsValidGameName(String gameName, Entity userEntity) {
...
+    Key gameKey = gameEntity.getKey();
+ 
+    datastore.put(gameEntity);
+
+    // getting the game id for the game id property
+    Entity entity = null;
+    try{
+      entity = datastore.get(gameKey);
+    }catch(EntityNotFoundException e){
+      return false;
```
hannahsalameh, line 75: why return false? maybe just run an error message instead and figure out how to deal w the fact that entity is equal to null

hannahsalameh: also I would put the key stuff below in the try block as well, since it can only run if the entity is found. otherwise u wouldnt want to put an empty entity in the datastore.

EmmaWenzel: The idea behind returning false is that if this doesn't work I'd want to break out of the function, because I don't want a game without an id to grab it with. That said, this should never happen, because what it's relying on is the entity that was created three lines above. The reason I have the try catch is because maven gets angry if I don't have it

src/test/java/com/google/sps/CreateGameTest.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  // Given an empty string for game name, the function should return false
+  @Test
+  public void emptyGameName() {
```
ofk-goog, line 54: My preference is to include the tested behavior in the name of the test method, e.g. `addingEmptyGameNameFails()`. (This is just a personal thing, though. I think Devon disagrees with me. Do whatever you want.)

src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -40,4 +43,42 @@ public boolean IsValidGameName(String gameName, Entity userEntity) {
     }
     return true;
   }
+
+  /**
+  * Creates a game entity 
+  */
+  public boolean createGameEntity(String gameName, DatastoreService datastore) {
+    
+    if(gameName == "" || datastore == null){
```
ofk-goog, line 52: If the `datastore` is null, that should be a fatal error. Or if not, it's at least worth logging.

src/main/java/com/google/sps/utils/GameUtils.java
EmmaWenzel: I used system.out to log the error; is that how you log an error? I saw some stuff about using an external logger like sentry or writing errors to a file, but I wasn't sure if I should be using either of those.

src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -40,4 +43,42 @@ public boolean IsValidGameName(String gameName, Entity userEntity) {
     }
     return true;
   }
+
+  /**
+  * Creates a game entity
```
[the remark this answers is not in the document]
EmmaWenzel: Got it, that's done.

src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -40,4 +43,42 @@ public boolean IsValidGameName(String gameName, Entity userEntity) {
     }
     return true;
   }
+
+  /**
+  * Creates a game entity 
+  */
+  public boolean createGameEntity(String gameName, DatastoreService datastore) {
```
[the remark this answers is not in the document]
EmmaWenzel: I made both methods static and revised the tests for that

src/test/java/com/google/sps/CreateGameTest.java
EmmaWenzel: I added "fails" to the ones that are supposed to fail

src/main/java/com/google/sps/utils/GameUtils.java
EmmaWenzel: I logged this with system.out as well

src/main/java/com/google/sps/utils/GameUtils.java
ofk-goog: Use `java.util.logging.Logger`
See the first Java code example here: https://cloud.google.com/appengine/docs/standard/java/logs
But you want to `log.severe` instead of `log.info`, because this is a "should never happen" error.

ofk-goog: Um. If it doesn't work, you may need to do something from this page: https://cloud.google.com/logging/docs/setup/java
I think you have some experience dealing with this sort of appengine config.

src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -45,12 +45,16 @@ public boolean IsValidGameName(String gameName, Entity userEntity) {
   }
 
   /**
-  * Creates a game entity 
+  * Creates a game entity and returns true if successful
```
ofk-goog, line 51: This comment is out of date

src/main/java/com/google/sps/utils/GameUtils.java
EmmaWenzel: The first one worked!

PR #23, opened 2020-07-14
created a class with helper functions to check if user has taken as q…

colemanliyah:
…uiz; also created tests for those functions. The tests check the parameters being fed into the functions.

### DIFF
--- a/src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
+++ b/src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
@@ -26,6 +26,7 @@ import com.google.appengine.api.datastore.Query.FilterPredicate;
  
 public final class QuizTimingPropertiesUtils {
  
+    //This function gets the the "quiz_timestamp" property of the entity that is fed into the function
     public Object getTimestampProperty(String entity, DatastoreService datastore) {
         try {
             Query query = new Query(entity);
@@ -43,6 +44,7 @@ public final class QuizTimingPropertiesUtils {
         }
     }
  
+    //This function checks if the user has taken the quiz yet by comparing their timestamp with the quiz's timestamp
     public Boolean userTookQuiz(String usersQuizTime, String CurrentQuizTime) {
         if(usersQuizTime.compareTo(CurrentQuizTime) > 0) {
             return true;
--- a/src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
+++ b/src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
@@ -1,0 +1,52 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+ 
+package com.google.sps;
+ 
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import java.util.List;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+ 
+public final class QuizTimingPropertiesUtils {
+ 
+    public Object getTimestampProperty(String entity, DatastoreService datastore) {
+        try {
+            Query query = new Query(entity);
+            PreparedQuery pq = datastore.prepare(query);
+ 
+            try {
+                Entity fetched_item = pq.asList(FetchOptions.Builder.withLimit(1)).get(0);
+                Object timestamp_of_fetcheditem = fetched_item.getProperty("quiz_timestamp");
+                return timestamp_of_fetcheditem;
+            } catch (IndexOutOfBoundsException e) {
+                return null;
+            }
+        } catch(NullPointerException e) {
+            return null;
+        }
+    }
+ 
+    public Boolean userTookQuiz(String usersQuizTime, String CurrentQuizTime) {
+        if(usersQuizTime.compareTo(CurrentQuizTime) > 0) {
+            return true;
+        } 
+        return false;
+    }
+}
--- a/src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
+++ b/src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
@@ -28,25 +28,23 @@ public final class QuizTimingPropertiesUtils {
  
     //This function gets the the "quiz_timestamp" property of the entity that is fed into the function
     public Object getTimestampProperty(String entity, DatastoreService datastore) {
+        Query query = new Query(entity);
+        PreparedQuery pq;
         try {
-            Query query = new Query(entity);
-            PreparedQuery pq = datastore.prepare(query);
- 
-            try {
-                Entity fetched_item = pq.asList(FetchOptions.Builder.withLimit(1)).get(0);
-                Object timestamp_of_fetcheditem = fetched_item.getProperty("quiz_timestamp");
-                return timestamp_of_fetcheditem;
-            } catch (IndexOutOfBoundsException e) {
-                return null;
-            }
+            pq = datastore.prepare(query);
         } catch(NullPointerException e) {
             return null;
         }
+        if(pq.asList(FetchOptions.Builder.withLimit(1)).size() > 0) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.appengine.api.datastore.Query;
+import java.util.List;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+ 
+public final class QuizTimingPropertiesUtils {
+ 
+    public Object getTimestampProperty(String entity, DatastoreService datastore) {
```
EmmaWenzel, line 29: Could you add a sentence or two of documentation for these functions? I'm a little confused on what they're doing

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -0,0 +1,84 @@
...
+        game = new Entity("game");
+    }
+ 
+    @After
+    public void tearDown() {
+        helper.tearDown();
+    }
+ 
+    @Test
+    public void nullUserTimeStamp() {
```
EmmaWenzel, line 55: could you add a bit of explanation for each of these tests too?

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+ 
+    @Test
+    public void noDatastore() {
+        Object actual = timing_properties_test.getTimestampProperty("user", null);
+        Assert.assertEquals(null, actual);
+    }
+ 
+    @Test
+    public void emptyStrings() {
```
EmmaWenzel, line 76: You don't have any tests that result in success (not null or true), I don't know if those are necessary but my instinct is that they'll be helpful

src/test/java/com/google/sps/TimingPropertiesTest.java
colemanliyah: Got it!

src/test/java/com/google/sps/TimingPropertiesTest.java
colemanliyah: Okay, I added two tests that check for a positive case

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
colemanliyah: Yes, i added some comments

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+ 
+public final class QuizTimingPropertiesUtils {
+ 
+    //This function gets the the "quiz_timestamp" property of the entity that is fed into the function
+    public Object getTimestampProperty(String entity, DatastoreService datastore) {
+        try {
+            Query query = new Query(entity);
+            PreparedQuery pq = datastore.prepare(query);
```
devonwasson, line 33: Typically, we only want to have one operation in a try catch block. In this case, making a new query object likely won't result in an error so we can put that outside the try block and only have the PreparedQuery in the try block.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -0,0 +1,54 @@
...
+ 
+            try {
```
devonwasson, line 35: We should try to avoid having this try catch block be inside another try catch block. Ideally, we can just try catch making the prepared query, then try catch this right afterwards, rather than doing it in a nested fashion.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -0,0 +1,54 @@
...
+                Entity fetched_item = pq.asList(FetchOptions.Builder.withLimit(1)).get(0);
```
devonwasson, line 36: .get(0) is the only thing that can cause the IndexOutOfBoundsException, so this is the only line that should be in the try/catch. However, we can avoid having to do a try/catch all together by testing if pq.asList(FetchOptions.Builder.withLimit(1)).size > 0 before trying to call .get(0).

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -0,0 +1,54 @@
...
+                Object timestamp_of_fetcheditem = fetched_item.getProperty("quiz_timestamp");
```
devonwasson, line 37: What specific type is timestamp_of_fetcheditem? We want to be more specific than using the generic "Object" class.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -0,0 +1,54 @@
...
+                return timestamp_of_fetcheditem;
```
devonwasson, line 38: We can combine this line with the above line, just directly returning fetched_item.getProperty("quiz_timestamp"); given the way this code is written. However, do we have any test cases to test if fetched_item does not have the property "quiz_timestamp" ?

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -0,0 +1,54 @@
...
+            } catch (IndexOutOfBoundsException e) {
+                return null;
+            }
+        } catch(NullPointerException e) {
+            return null;
+        }
+    }
+ 
+    //This function checks if the user has taken the quiz yet by comparing their timestamp with the quiz's timestamp
+    public Boolean userTookQuiz(String usersQuizTime, String CurrentQuizTime) {
```
devonwasson, line 48: CurrentQuizTime should start with a lower case c

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.runners.JUnit4;
+import com.google.sps.QuizTimingPropertiesUtils;
+ 
+@RunWith(JUnit4.class)
+public final class TimingPropertiesTest {
+  
+    private DatastoreService datastore;
+    private QuizTimingPropertiesUtils timing_properties_test;
+    public Entity user;
+    public Entity game;
```
devonwasson, line 36: We'll want to make new entities in every test. Having them be global variables can cause issues with the tests. What if I set a property in the first test but that property conflicts with the expected output of the second test?

This also means that you should remove their declaration from the setUp() as well and just do it explicitly in each test function.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
colemanliyah: I got rid of the timestamp_of_fetcheditem variable like you suggested, should I still be more specific about the object I am returning? I just do it this way because it is easy to then convert to a string later

src/test/java/com/google/sps/TimingPropertiesTest.java
colemanliyah: The properties I set to these entitys are only set within tests so that they don't cross over, do you still think I should define them for each test? I kept them as global variables only to reduce the code

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -75,6 +75,13 @@ public void noDatastore() {
         Assert.assertEquals(null, actual);
     }
 
+    @Test
+    //If the entity doesn't "timestamp" property then null should be returned
```
devonwasson, line 79: I think this sentence is missing a word.

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -75,6 +75,13 @@ public void noDatastore() {
...
+    public void noTimeStampProperty() {
+        Object actual = timing_properties_test.getTimestampProperty("user", datastore);
```
devonwasson, line 81: Just realized but we should be accepting the specific object type and not just Object here as well.

src/test/java/com/google/sps/TimingPropertiesTest.java
devonwasson: We also would probably want to have DatastoreService datastore and QuizTimingPropertiesUtils timing_properties_test each be specific to each test. We don't want to share variables across multiple tests.

helper.setUp() should stay in the setUp() function, but everything else should be repeated on a per-test basis.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
colemanliyah: I do not, when I print the type it says it is a "java.lang.Object"

devonwasson: https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/PropertyContainer.html#getProperty-java.lang.String-

Well this is really annoying but I guess it does just return Object... That's... just the worst... but nothing you can do about that, so you're good to keep everything as "Object" lol. Sorry about the non-relevant advice!

PR #24, opened 2020-07-14
change the way the UserUtil fxn works

hannahsalameh:
made it so QueryByUserId was more abstract and can query from any place in the datastore instead of it being hardcoded in. I also changed the name of it from QueryByUserId to FilteredQuery and changed all the previous tests to now run with the new name and inputs.


src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -26,20 +26,31 @@
...
+    * This method always returns immediately, whether or not the 
+    * datastore object exists. If the object does not exist, the entity 
+    * returned is null.
+    * 
+    * @param  entityName           the name of an entity class in the datastore (ex.user, game)
+    * @param  entityPropertyTitle  the name of a specific property that the entity class has (usually sessionID or userID)
+    * @param  entityPropertyValue  the value of the property that is filtered for
+    * @return                      a single entity that has the same value for the property in the parameter 
+    */
+  public Entity FilteredQuery(String entityName, String entityPropertyTitle, String entityPropertyValue, DatastoreService datastore) {
```
devonwasson, line 43: Function names should typically be nouns, so something like "CreateFilteredQuery" or "BuildFilteredQuery". That being said, considering this function returns an Entity, we should probably not have the word Query in the function name, and instead call it something like "GetEntityFromDatastore"

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -26,20 +26,31 @@
...
+    if(entityPropertyValue == "" || entityName == "" || entityPropertyTitle == "" || datastore == null){
         return null;
     }
 
-    Filter getCorrectUser = new FilterPredicate("userID", FilterOperator.EQUAL, givenUser);
-    Query query = new Query("Game").setFilter(getCorrectUser);
-    PreparedQuery results = datastore.prepare(query);
-    
-    List<Entity> resultsList = results.asList(FetchOptions.Builder.withLimit(1));
+    Filter getCorrectUser = new FilterPredicate(entityPropertyTitle, FilterOperator.EQUAL, entityPropertyValue);
```
devonwasson, line 49: getCorrectUser sounds like a function name and not a variable name. This should be something like "userQueryFilter" or just "queryFilter"

hannahsalameh: changed!

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -14,6 +14,13 @@
 
 package com.google.sps;
 
+
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
```
devonwasson, line 23: Are any of these new imports used?

src/test/java/com/google/sps/GetEntityFromDatastoreTest.java
```diff
@@ -0,0 +1,155 @@
...
+import org.junit.Before;
+import org.junit.After;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import com.google.sps.UserUtils;
+
+/** tests the IsValidGameName function */
+@RunWith(JUnit4.class)
+public final class GetEntityFromDatastoreTest {
```
devonwasson, line 32: I think you just meant to change the name of the existing test file you already have but you accidentally created a copy of it instead, so now there are two copies of the same test code in this PR, just with different names.

src/main/java/com/google/sps/utils/UserUtils.java
hannahsalameh: You need some of them for the cookie object, ill check which and take some out

PR #25, opened 2020-07-15
Add helper function to add a given user to the game entity and create…

EmmaWenzel:
… a score entity for that game for that user. When a user is added to a game, they are given a score for that specific game which is stored in a score entity that has the users id and the game ids to query it from. The tests cover null/empty parameters and a user entity without a list of games (this should never happen, so the error is logged).

### DIFF
--- a/src/main/java/com/google/sps/utils/GameUtils.java
+++ b/src/main/java/com/google/sps/utils/GameUtils.java
@@ -89,4 +89,43 @@ public final class GameUtils {
  
     return entity;
   }
+
+  /**
+  * add user to user list in game entity, and creates a score entity for the given user and game
+  */
+  public static boolean addUserToGame(String userId, String userName, Entity gameEntity, DatastoreService datastore) {
+ 
+    if(userId == "" || userName == ""){
+      return false;
+    }
+
+    if(gameEntity == null || datastore == null){
+      log.severe("null entity or datastore"); 
+      return false;
+    }
+ 
+    ArrayList<String> userIds = (ArrayList<String>) gameEntity.getProperty("userIds");
+    if(userIds == null){
+      log.severe("empty user id list");
+      return false;
+    }
+ 
+    // create score entity for the user for this game (starts as 0)
+    int score = 0;
+    Entity scoreEntity = new Entity("Score");
+    scoreEntity.setProperty("userId", userId);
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -89,4 +89,43 @@ public static Entity createGameEntity(String gameName, DatastoreService datastor
...
+      return false;
+    }
+
+    if(gameEntity == null || datastore == null){
+      log.severe("null entity or datastore"); 
+      return false;
+    }
+ 
+    ArrayList<String> userIds = (ArrayList<String>) gameEntity.getProperty("userIds");
+    if(userIds == null){
```
colemanliyah, line 108: What if the size of userIds is 0 because no users have been added yet. Do you still want to return false then?

EmmaWenzel: No, if the size of user ids is 0 in some cases you still want to add the user (so like if the game was just created) this is what the last test case is testing (because nothing is added to userIds after it's initialized, it's size is 0)

    ArrayList<String> userIds = new ArrayList<>();
    gameEntity.setProperty("userIds", userIds);

    boolean actual = GameUtils.addUserToGame("user1", "1", gameEntity, datastore);

    Assert.assertEquals(true, actual);

src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -89,4 +89,43 @@ public static Entity createGameEntity(String gameName, DatastoreService datastor
...
+      log.severe("empty user id list");
+      return false;
+    }
+ 
+    // create score entity for the user for this game (starts as 0)
+    int score = 0;
+    Entity scoreEntity = new Entity("Score");
+    scoreEntity.setProperty("userId", userId);
+    scoreEntity.setProperty("userName", userName);
```
ofk-goog, line 117: Are you sure you need to store the userName in the Score entity? I recommend holding off on this until you decide you need it. Which I suspect you won't, because you can use the userId to look up the userName.

(My point is that if you avoid storing the same information in multiple places, the system will be simpler and you'll run into fewer problems later.)

EmmaWenzel: That's a good point, I took it out

PR #26, opened 2020-07-15
create helper function to find a specific cookie

hannahsalameh:
Created a helper function to query a specific cookie given the cookie's name. Wrote tests for this also

### DIFF
--- a/src/main/java/com/google/sps/utils/CookieUtils.java
+++ b/src/main/java/com/google/sps/utils/CookieUtils.java
@@ -20,10 +20,10 @@ import javax.servlet.http.*;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-
+import java.util.logging.Logger;
 
 public final class CookieUtils {
-
+     private static final Logger log = Logger.getLogger(UserUtils.class.getName());
         /**
     * Returns a single Cookie object that can then be used in a servlet. 
     * The name arguement must be the name of a cookie active
@@ -37,6 +37,7 @@ public final class CookieUtils {
 
     public static Cookie getCookieFromName(Cookie cookies[], String name){
         if(name == "" || cookies.length == 0){
+            log.severe("bad inputs");
             return null;
         }
         for(int i = 0; i < cookies.length; i++){
--- a/src/main/java/com/google/sps/utils/CookieUtils.java
+++ b/src/main/java/com/google/sps/utils/CookieUtils.java
@@ -29,16 +29,13 @@ public final class CookieUtils {
     * The name arguement must be the name of a cookie active
     * on the front-end for a Cookie to be retrieved.
     * <p>
-    * This method always returns immediately, whether or not the 
-    * datastore object exists. If the object does not exist, the cookie 
-    * returned is null.
     * 
     * @param  cookies  an array of cookies (usually all of the cookies on the front end)
     * @param  name     the name of a specific cookie
     * @return          the cookie w the name passed in as a parameter
     */
 
-    public Cookie getCookieGivenName(Cookie cookies[], String name){
+    public Cookie getCookieFromName(Cookie cookies[], String name){
         if(name == "" || cookies.length == 0){
             return null;
         }
--- a/src/main/java/com/google/sps/utils/CookieUtils.java
+++ b/src/main/java/com/google/sps/utils/CookieUtils.java
@@ -1,0 +1,53 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+
+public final class CookieUtils {
+
+        /**
+    * Returns a single Cookie object that can then be used in a servlet. 
+    * The name arguement must be the name of a cookie active
+    * on the front-end for a Cookie to be retrieved.
+    * <p>
+    * This method always returns immediately, whether or not the 
+    * datastore object exists. If the object does not exist, the cookie 
+    * returned is null.
+    * 
+    * @param  cookies  an array of cookies (usually all of the cookies on the front end)
+    * @param  name     the name of a specific cookie
+    * @return          the cookie w the name passed in as a parameter
+    */
+
+    public Cookie getCookieGivenName(Cookie cookies[], String name){
+        if(name == "" || cookies.length == 0){
+            return null;
+        }
+        for(int i = 0; i < cookies.length; i++){
+            Cookie cookie = cookies[i];
+            if(cookie.getName() == name){
+                return cookie;
+            }
+        }
+        return null;
+    }
+}
--- a/src/main/java/com/google/sps/utils/CookieUtils.java
+++ b/src/main/java/com/google/sps/utils/CookieUtils.java
@@ -36,10 +36,15 @@ public final class CookieUtils {
     */
 
     public static Cookie getCookieFromName(Cookie cookies[], String name){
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/utils/CookieUtils.java
```diff
@@ -0,0 +1,53 @@
...
+public final class CookieUtils {
+
+        /**
+    * Returns a single Cookie object that can then be used in a servlet. 
+    * The name arguement must be the name of a cookie active
+    * on the front-end for a Cookie to be retrieved.
+    * <p>
+    * This method always returns immediately, whether or not the 
+    * datastore object exists. If the object does not exist, the cookie 
+    * returned is null.
```
devonwasson, line 34: We don't use a datastore in this method, so what does this comment refer to?

src/main/java/com/google/sps/utils/CookieUtils.java
```diff
@@ -0,0 +1,53 @@
...
+    * 
+    * @param  cookies  an array of cookies (usually all of the cookies on the front end)
+    * @param  name     the name of a specific cookie
+    * @return          the cookie w the name passed in as a parameter
+    */
+
+    public Cookie getCookieGivenName(Cookie cookies[], String name){
```
devonwasson, line 41: super nit picky here but "GetCookieFromName" might be clearer.

src/test/java/com/google/sps/GetCookieGivenNameTest.java
```diff
@@ -0,0 +1,118 @@
...
+import javax.servlet.http.HttpServletResponse;
+import com.google.sps.CookieUtils; 
+
+@RunWith(JUnit4.class)
+public final class GetCookieGivenNameTest {
+    private CookieUtils cookieUtils;
+
+  @Before
+  public void setUp() {
+    cookieUtils = new CookieUtils();
```
devonwasson, line 29: We actually might want to have a separate utility function declared in each test method. It's usually best to avoid sharing objects across multiple tests.

src/test/java/com/google/sps/GetCookieGivenNameTest.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+   @After
+  public void tearDown() {
+  }
```
devonwasson, line 34: If nothing is in the before and after methods, we can remove them.

src/test/java/com/google/sps/GetCookieGivenNameTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+  //given a list of cookies, it should find the one with the mentioned name.
+  @Test
+  public void findCookieByName() {
+        Cookie cookie1 = new Cookie("hello", "got it");
+        Cookie cookie2 = new Cookie("hi", "not this one");
+        Cookie cookie3 = new Cookie("heyyy", "not this one either");
```
devonwasson, line 41: Values that are more descriptive might be helpful, like "name1" and "value1". They're very generic but they tell a better story. (This applies to all other cookies in this test)

src/test/java/com/google/sps/GetCookieGivenNameTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+        Cookie cookieArray[] = new Cookie[]{cookie1,cookie2,cookie3};
+
+        Cookie actual = cookieUtils.getCookieGivenName(cookieArray, "hello");
+        Cookie expected = cookie1;
+
+         Assert.assertEquals(expected, actual);     
+  }
+
+//given a list of cookies and a name that isn't in the list, returns null
```
devonwasson, line 51: This should tell us what we're testing rather than specific behavior. So it should read something like "Tests when no cookie with the given name exists." This advice applies to all other test methods.

src/test/java/com/google/sps/GetCookieGivenNameTest.java
```diff
@@ -0,0 +1,118 @@
...
+        Cookie cookie1 = new Cookie("hello", "got it");
+        Cookie cookie2 = new Cookie("hi", "not this one");
+        Cookie cookie3 = new Cookie("heyyy", "not this one either");
+
+        Cookie cookieArray[] = new Cookie[]{cookie1,cookie2,cookie3};
+
+        Cookie actual = cookieUtils.getCookieGivenName(cookieArray, "hello");
+        Cookie expected = cookie1;
+
+         Assert.assertEquals(expected, actual);
```
devonwasson, line 48: All of your assert lines are one space indented more than the others. They should be at the same level of indentation.

src/main/java/com/google/sps/utils/CookieUtils.java
hannahsalameh: oh sorry this is copy pasted and i forgot to take it out

src/main/java/com/google/sps/utils/CookieUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.servlet.*;
+import javax.servlet.http.*;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+
+public final class CookieUtils {
+
+        /**
```
colemanliyah, line 27: I would add the Logger package that Emma mentioned here in the imports and then create an object of it and write at least one severe for your function

src/test/java/com/google/sps/CookieUtilsTest.java
```diff
@@ -0,0 +1,105 @@
...
+        Cookie cookie1 = new Cookie("name1", "value1");
+        Cookie cookie2 = new Cookie("name2", "value2");
+        Cookie cookie3 = new Cookie("name3", "value3");
+
+        Cookie cookieArray[] = new Cookie[]{cookie1,cookie2,cookie3};
+
+        Cookie actual = CookieUtils.getCookieGivenName(cookieArray, "name1");
+        Cookie expected = cookie1;
+
+        Assert.assertEquals(expected, actual);
```
colemanliyah, line 35: I would just do Asser.assertEquals(cooke1, actual) and get rid of line 33

src/test/java/com/google/sps/CookieUtilsTest.java
```diff
@@ -0,0 +1,105 @@
...
+        Cookie cookie1 = new Cookie("name1", "value1");
+        Cookie cookie2 = new Cookie("name2", "value2");
+        Cookie cookie3 = new Cookie("name3", "value3");
+
+        Cookie cookieArray[] = new Cookie[]{cookie1,cookie2,cookie3};
+
+        Cookie actual = CookieUtils.getCookieGivenName(cookieArray, "noname");
+        Cookie expected = null;
+
+        Assert.assertEquals(expected, actual);
```
colemanliyah, line 50: Same comment as above

src/main/java/com/google/sps/utils/CookieUtils.java
```diff
@@ -20,26 +20,24 @@
...
     * 
     * @param  cookies  an array of cookies (usually all of the cookies on the front end)
     * @param  name     the name of a specific cookie
     * @return          the cookie w the name passed in as a parameter
     */
 
-    public Cookie getCookieGivenName(Cookie cookies[], String name){
+    public static Cookie getCookieFromName(Cookie cookies[], String name){
         if(name == "" || cookies.length == 0){
+            log.severe("bad inputs");
```
devonwasson, line 40: Please see notes about logging from today's stand up meeting.

src/main/java/com/google/sps/utils/CookieUtils.java
```diff
@@ -0,0 +1,56 @@
...
+    * @return          the cookie w the name passed in as a parameter
+    */
+
+    public static Cookie getCookieFromName(Cookie cookies[], String name){
+        if(name == ""){
+            log.severe("found null search parameter while trying to run function getCookieFromName");
+            return null;
+        }
+        if(cookies.length == 0){
+             log.severe("found no cookies while trying to run function getCookieFromName");
```
devonwasson, line 44: This could be a little more helpful with something like "List of cookies is empty when searching for cookie with name: <name> in getCookieFromName()" 

Or however you want to do it, but we should include the name at this point (because we can ensure that the name is not empty here).

src/main/java/com/google/sps/utils/CookieUtils.java
```diff
@@ -0,0 +1,56 @@
...
+    * <p>
+    * 
+    * @param  cookies  an array of cookies (usually all of the cookies on the front end)
+    * @param  name     the name of a specific cookie
+    * @return          the cookie w the name passed in as a parameter
+    */
+
+    public static Cookie getCookieFromName(Cookie cookies[], String name){
+        if(name == ""){
+            log.severe("found null search parameter while trying to run function getCookieFromName");
```
devonwasson, line 40: The string isn't null, it's empty. It's a subtle difference but worth capturing. So this should be changed to something like "Empty string name passed to getCookieFromName()"

PR #27, opened 2020-07-15
Get Entity from Session ID

hannahsalameh:
created a helper fxn to get an entity from a session ID. depends on the cookieUtils PR i just made


src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -65,4 +65,33 @@ public Entity getEntityFromDatastore(String entityName, String entityPropertyTit
     Entity entity = resultsList.get(0);
     return entity;
   }
+
+   
+   /**
+    * Returns a single Entity object that can then be used in a servlet.
```
EmmaWenzel, line 71: u can take out "that can then be used in a servlet". I think that might count as putting stuff about how the servlet should be used as opposed to just what it does in the description. I would also specify that it's a user entity

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -65,4 +65,33 @@ public Entity getEntityFromDatastore(String entityName, String entityPropertyTit
...
+    * returned is null.
+    * 
+    * @param  cookies    an array of cookies (usually all cookies on front end)
+    * @param  datastore  the database where entities are stored
+    * @return            a single entity that has the cookie name/value pair as a property
+    */
+ 
+  public Entity getUserFromCookie(Cookie cookies[], DatastoreService datastore){
+      if(datastore == null){
+          return null;
```
EmmaWenzel, line 85: I'd log this, you can see the screen shot I sent in the chat it should be exactly the same

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -65,4 +65,33 @@ public Entity getEntityFromDatastore(String entityName, String entityPropertyTit
...
+      }
+      String sessionIDName = "SessionID";
+      CookieUtils cookieUtils = new CookieUtils();
+      Cookie cookie = cookieUtils.getCookieGivenName(cookies, sessionIDName); //returns null if it doesn't exist
```
EmmaWenzel, line 89: If you make this function static (like public static Cookie getCookieName(...)) you can just call this with 
Cookie cookie = CookieUtils.getCookieGivenName(cookies, sessionIDName)
instead of 
CookieUtils cookieUtils = new CookieUtils();
Cookie cookie = cookieUtils.getCookieGivenName(cookies, sessionIDName);

you can also do this for this function so you can use it elsewhere the same

PR #28, opened 2020-07-15
Liyahs working branch

colemanliyah:
I added the functionality of checking if new quiz question needs to be generated based if it is a new day. This functionality is in helper functions. I also created the tests to handle valid and invalid parameters


src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -18,6 +18,41 @@
...
+        "Which plant represents your best friend?",
+        "Which plant will most likely impress your friends and family?",
+        "Which plant would look the best inside as a houseplant?",
+        "Which plant would look the best outside in a garden?",
+        "Which plant would you give as a gift?"
+    ));
+
 import java.util.List;
 import com.google.appengine.api.datastore.FetchOptions;
 import com.google.appengine.api.datastore.Query.Filter;
```
hannahsalameh, line 58: u might wanna move these imports so they're before the class declaration

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -138,5 +138,55 @@ public void validParameters_for_userTookQuiz() {
...
+ 
+        Boolean actual = timing_properties_test.newDayNewQuiz(null);
+        Assert.assertEquals(null, actual);
+    }
  
+    @Test 
+    //Checks if newDayNewQuiz works especially with being dependent on the getTimestampProperty
+    public void validParameter_for_newDayNewQuiz() {
+        QuizTimingPropertiesUtils timing_properties_test = new QuizTimingPropertiesUtils();
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
hannahsalameh, line 155: I think that datastore and the timing properties test can be declared once for all the tests, its only instances of the class you are testing that would have to be declared in each separate test

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -138,5 +138,55 @@ public void validParameters_for_userTookQuiz() {
...
+        datastore.put(user);
+ 
+        Object user_quiz_time = timing_properties_test.getTimestampProperty("user", datastore);
+ 
+        Boolean actual = timing_properties_test.newDayNewQuiz(user_quiz_time);
+        Assert.assertEquals(true, actual);
+    }
+ 
+    @Test 
+    //If the Entity parameter is null then the getNewQuestion function should return null
```
hannahsalameh, line 168: Comments shouldn't say the expected value, but instead just describe the inputs (this is a problem on one of my other PRs)
Ex. Test where the entity parameter is null

src/test/java/com/google/sps/TimingPropertiesTest.java
colemanliyah: Devon suggested I create them for each test which is why I don't have them as global variables for the whole class

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
colemanliyah: That's weird, these imports were suppose to be deleted...let me know if you still see them in the new commit

src/test/java/com/google/sps/TimingPropertiesTest.java
colemanliyah: Changed them! Let me know if the new ones are better

src/test/java/com/google/sps/TimingPropertiesTest.java
hannahsalameh: ok cool thats fine then

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -49,4 +78,44 @@ public Boolean userTookQuiz(String usersQuizTime, String currentQuizTime) {
...
+ 
+        String today_date = DateFormat.getDateInstance().format(new Date());
+        if(today_date.compareTo(quiz_date) > 0) {
+            return true;
+        }
+        return false;
+    }
+    
+    //This function gets a new quiz question if the quiz is outdated
+    public Object getNewQuestion(Entity game_entity, DatastoreService datastore) {
```
ofk-goog, line 100: In my experience, `Object` is never a useful return type. Please make the return type more specific.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -49,4 +78,44 @@ public Boolean userTookQuiz(String usersQuizTime, String currentQuizTime) {
...
+        } catch(IllegalArgumentException e) {
+            log.severe("Given a null Parameter");
+            return null;
+        }
+ 
+        String today_date = DateFormat.getDateInstance().format(new Date());
+        if(today_date.compareTo(quiz_date) > 0) {
+            return true;
+        }
+        return false;
```
ofk-goog, line 96: `return today_date.compareTo(quiz_date) > 0;`

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -49,4 +78,44 @@ public Boolean userTookQuiz(String usersQuizTime, String currentQuizTime) {
         } 
         return false;
     }
+
+    //This function checks to see if the quiz is outdated
+     public Boolean newDayNewQuiz(Object current_quiz_time) {
```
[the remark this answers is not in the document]
colemanliyah: Devon mentioned this as well but I mentioned that what is returned by getTimestampProperty() is a java object, it's type does not print out to be more specific than that

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -33,12 +60,14 @@ public Object getTimestampProperty(String entity, DatastoreService datastore) {
         try {
             pq = datastore.prepare(query);
         } catch(NullPointerException e) {
+            log.severe("Null datastore");
             return null;
```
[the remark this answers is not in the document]
colemanliyah: Does this mean that instead of the return statement I should just have the log.severe() part? Won't that make the tests fail though?

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -33,12 +60,14 @@ public Object getTimestampProperty(String entity, DatastoreService datastore) {
...
         }
         if(pq.asList(FetchOptions.Builder.withLimit(1)).size() > 0) {
             Entity fetched_item = pq.asList(FetchOptions.Builder.withLimit(1)).get(0);
             return fetched_item.getProperty("quiz_timestamp");
         } 
+        log.severe("Zero Items Quered");
```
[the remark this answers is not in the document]
colemanliyah: I have not been able to get this syntax to work yet... is there a way I can do what you are suggesting but still keep it as log.severe(""); ?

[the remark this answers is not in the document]
ofk-goog: Not as far as I know. Are you getting a compiler error or a runtime error? What's the message?

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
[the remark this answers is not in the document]
ofk-goog: No, my suggestion would be to not catch the NullPointerException in the first place, and instead just let the program crash. But I think Java makes it difficult to do this, so it might not be worth worrying about for now.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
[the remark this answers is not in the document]
colemanliyah: It says it cannot find the symbol "Level" If I get rid of Level.SEVERE and just put SEVERE then it says it can't find symbol "SEVERE"

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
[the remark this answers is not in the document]
ofk-goog: You've updated `getTimestampProperty()` to return a `Long` (good job), so now please make this parameter a `Long` as well.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -65,22 +64,19 @@ public Object getTimestampProperty(String entity, DatastoreService datastore) {
...
-        return false;
+        return (usersQuizTime.compareTo(currentQuizTime) > 0);
     }
 
     //This function checks to see if the quiz is outdated
-     public Boolean newDayNewQuiz(Object current_quiz_time) {
+     public Boolean isQuizOutdated(Object current_quiz_time) {
         String quiz_date;
         try {
             quiz_date = DateFormat.getDateInstance().format(current_quiz_time);
```
ofk-goog, line 82: The behavior of this code depends on the configured time zone of the computer the code runs on. That can lead to problems when your code is running in the cloud. I recommend either specifying a time zone when converting the `Long` to a date, or else comparing the timestamps without converting them to dates.

colemanliyah: Okay I did the latter suggestion, let me know if that is what you meant

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -67,28 +68,29 @@ public Long getTimestampProperty(String entity, DatastoreService datastore) {
...
 
     //This function checks to see if the quiz is outdated
-     public Boolean isQuizOutdated(Long current_quiz_time) {
-        String quiz_date;
+     public Boolean isQuizOutdated(Long current_quiz_time) {    
+        Long today_date = System.currentTimeMillis();
         try {
-            quiz_date = DateFormat.getDateInstance().format(current_quiz_time);
-        } catch(IllegalArgumentException e) {
+            return today_date > current_quiz_time;
```
ofk-goog, line 89: What is current_quiz_time? I thought it was the time the quiz was created, which would be in the past. If so, this will always return `true`.

`System.currentTimeMillis() - current_quiz_time` would give the age of the quiz in milliseconds

colemanliyah: it is the time the quiz is created but everyday it gets set to a new time. So if today's date is greater than the quiz time then we know we have to create a new quiz question and subsequently update the time of the quiz

ofk-goog: `System.currentTimeMillis()` is not today's date -- it's the number of milliseconds since Jan 1 1970 at midnight UTC.

PR #29, opened 2020-07-15
Add user helper methods

EmmaWenzel:
Added two methods to user utils: addGameToUser adds the game id to the list of game ids a user has. The tests make sure the game id isn't empty, the user and datastore aren't null, and an null list for game ids isn't gotten. addBlobKey adds a blobKey property to a user entity. The tests test for cases where there's an empty blobkey or null user entity or datastore passed in.


src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -62,4 +66,46 @@ public Entity getEntityFromDatastore(String entityName, String entityPropertyTit
...
   }
+
+  /**
+  * Adds a game id to a user's list of games
+  */
+  public static boolean addGameToUser(Entity userEntity, DatastoreService datastore, String gameId) {
+ 
+    if(gameId == ""){
+        return false;
+    }
```
colemanliyah, line 77: Every spot where the function can return false is not what we want right? So should there be a log.severe() in each of these places?

src/test/java/com/google/sps/UserUtilTest.java
```diff
@@ -159,4 +160,105 @@ public void nullUserEntityTitle(){
       Assert.assertEquals(expected, actual);
   }
 
+    // Given an empty string for game id, addGameToUser should return false
+  @Test
```
colemanliyah, line 164: Hannah mentioned that for comments explaining test, you dont need to put what you except the outcome to be you can just say what it is testing. For instance here it would be "Test empty string as paramter for game id" or something similar

src/test/java/com/google/sps/UserUtilTest.java
```diff
@@ -159,4 +160,105 @@ public void nullUserEntityTitle(){
...
+  public void emptyGameIdFails() {
+
+    Entity userEntity = new Entity("User");
```
colemanliyah, line 167: Also I think this line will be declared as "user" with lower case "u"

EmmaWenzel: yep, done!

src/test/java/com/google/sps/UserUtilTest.java
EmmaWenzel: got it, I changed all of the comments to reflect that

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: I added a few more

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -62,4 +66,54 @@ public Entity getEntityFromDatastore(String entityName, String entityPropertyTit
...
+  */
+  public static boolean addGameToUser(Entity userEntity, DatastoreService datastore, String gameId) {
+ 
+    if(gameId == ""){
+        log.severe("no gameId");
+        return false;
+    }
+
+    if(userEntity == null || datastore == null){
+        log.severe("null datastore or user entity");
```
devonwasson, line 81: This isn't a super helpful logging statement. If I read this, I'll still have questions (specifically, "well, which one is it?"). This can be improved by having one if statement per check and logging at each step.

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -62,4 +66,54 @@ public Entity getEntityFromDatastore(String entityName, String entityPropertyTit
...
+        return false;
+    }
+ 
+    ArrayList<String> games = (ArrayList<String>) userEntity.getProperty("games");
+ 
+    if(games == null){
+        log.severe("no game list");
```
devonwasson, line 88: Here we would want to log the username as well just so we know who we're talking about. However, if a user having no games is an expected failure mode, we might not want to log at all.

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -62,4 +66,54 @@ public Entity getEntityFromDatastore(String entityName, String entityPropertyTit
...
+    userEntity.setProperty("games", games);
+    datastore.put(userEntity);
+ 
+    return true;
+  }
+
+  /**
+  * adds a photo to the user entity
+  */
+  public static boolean addBlobKey(String blobKey, Entity userEntity, DatastoreService datastore) {
```
devonwasson, line 102: Similar comments about logging to those in the method above.

src/test/java/com/google/sps/UserUtilTest.java
```diff
@@ -19,17 +19,18 @@
...
 import org.junit.runners.JUnit4;
-import com.google.sps.UserUtils;
+import com.google.sps.utils.UserUtils;
 
 /** tests the IsValidGameName function */
 @RunWith(JUnit4.class)
-public final class GetEntityFromDatastoreTest {
+public final class UserUtilTest {
   
   private DatastoreService datastore;
```
devonwasson, line 35: I might have just missed this before, but we shouldn't have global mutable variables in a test like this. If I edit this variable in one test, the state will persist to another test, potentially making for unexpected failures. It's best to just create a new datastore in each test.

src/test/java/com/google/sps/UserUtilTest.java
```diff
@@ -19,17 +19,18 @@
...
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
-import com.google.sps.UserUtils;
+import com.google.sps.utils.UserUtils;
 
 /** tests the IsValidGameName function */
 @RunWith(JUnit4.class)
-public final class GetEntityFromDatastoreTest {
+public final class UserUtilTest {
```
devonwasson, line 33: Now that we are testing multiple functions in one test class, we should do one of two things:

Make multiple classes within this method, each class being named the method it is testing (whoops, I might have recommended otherwise before, but I'm reversing course now that we're testing multiple methods in a single class lol).

Prepend each method with the name of the function it is testing. 

I prefer the former option but this is my fault for recommending naming the class after the util and putting all method test cases in one test class. I think it makes more sense to split tests by method name into their own test classes and then have one test file per utility class.

devonwasson: disregard this message, I'm sending something else in the group chat.

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -62,4 +66,54 @@ public Entity getEntityFromDatastore(String entityName, String entityPropertyTit
...
     return entity;
   }
+
+  /**
+  * Adds a game id to a user's list of games
+  */
+  public static boolean addGameToUser(Entity userEntity, DatastoreService datastore, String gameId) {
+ 
+    if(gameId == ""){
+        log.severe("no gameId");
```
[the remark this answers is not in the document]
EmmaWenzel: Sounds good, I flipped that if statement around and made the log more specific

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: I split them up and made them more specific

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: Okay, I took this log out.

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: Sounds good, I made similar changes.

src/test/java/com/google/sps/UserUtilTest.java
EmmaWenzel: Okay, that's done for this file. I'll make another pull request to make these changes in the game utils file.

PR #30, opened 2020-07-16
Game util edits

EmmaWenzel:
Transfering the fixes from user utils into game utils; fix logging to be more specific, put all tests for game utils class in the same file, fix test names and comments, and remove global datastore variable.


src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -94,19 +98,21 @@ public static Entity createGameEntity(String gameName, DatastoreService datastor
...
+        return false;
+    }
+    if(datastore == null){
+        log.severe("found null datastore trying to game to user " + userId);
+        return false;
     }
  
     ArrayList<String> userIds = (ArrayList<String>) gameEntity.getProperty("userIds");
     if(userIds == null){
-      log.severe("empty user id list");
```
hannahsalameh, line 109: why did u remove the log here?

src/test/java/com/google/sps/GameUtilTest.java
```diff
@@ -0,0 +1,245 @@
...
+  public void setUp() {
+    helper.setUp();
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  //  Given a user has no games, any game name would be valid
```
hannahsalameh, line 53: Change ur comments to be more reflective of the test instead of the output

Ex. Test where a user has no games

src/test/java/com/google/sps/GameUtilTest.java
```diff
@@ -0,0 +1,245 @@
...
+  // The user has games, but the given game name is new
+  @Test
+  public void isValidGameNewGameName() {
+
+    Entity userEntity1 = new Entity("user");
+    ArrayList<String> gameNames1 = new ArrayList<>();
+    
+    gameNames1.add("game1");
+    gameNames1.add("game2");
+    gameNames1.add("Game3");
```
hannahsalameh, line 95: why is this one capital?

src/test/java/com/google/sps/GameUtilTest.java
EmmaWenzel: I think I understand what you mean, I changed all the test comments, let me know if they're still iffy.

src/test/java/com/google/sps/GameUtilTest.java
EmmaWenzel: oops missed that. It's changed.

src/main/java/com/google/sps/utils/GameUtils.java
EmmaWenzel: It's not a fatal error, it just means the list wasn't initialized, so that can be delt with after the function is called.

src/test/java/com/google/sps/GameUtilTest.java
```diff
@@ -0,0 +1,245 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import com.google.sps.utils.GameUtils;
+
+/** tests the game util functions */
+@RunWith(JUnit4.class)
+public final class GameUtilTest {
+
+  private static String userID1 = "1";
+  private static String userID2 = "2";
```
devonwasson, line 37: We can take or leave this. While it's usually good to define strings like this in classes, we typically don't need to in tests. It's usually more helpful to be as explicit as possible in test cases. However, I don't think it's awful here so we can probably leave it.

src/test/java/com/google/sps/GameUtilTest.java
devonwasson: Did you push the change? It still shows up as upper case for me.

src/test/java/com/google/sps/GameUtilTest.java
```diff
@@ -0,0 +1,245 @@
...
+  // The user has games, and the given name is a duplicate name with different capitalization
+  @Test
+  public void isValidGame_SameNameDifferentCapitalization() {
+
+    Entity userEntity1 = new Entity("user");
+    ArrayList<String> gameNames1 = new ArrayList<>();
+    
+    gameNames1.add("game1");
+    gameNames1.add("game2");
+    gameNames1.add("Game3");
```
devonwasson, line 113: same comment with capitalization

src/test/java/com/google/sps/GameUtilTest.java
EmmaWenzel: Makes more sense to take it out, so I did.

src/test/java/com/google/sps/GameUtilTest.java
EmmaWenzel: I did 1/3 apparently, those should all be fixed now.

PR #31, opened 2020-07-17
servlet for creating an upload url for plant images

EmmaWenzel:
This function is called on image upload; it creates a url and attaches it to the form so the browser can upload the file directly to blobstore. The serve image servlet obtains a url to serve the blob image.

### DIFF
--- a/src/main/java/com/google/sps/servlets/BlobstoreUploadUrlServlet.java
+++ b/src/main/java/com/google/sps/servlets/BlobstoreUploadUrlServlet.java
@@ -21,13 +21,11 @@ import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import com.google.sps.utils.SafeSearchUtils;
  
 /**
  * When the fetch() function requests the /blobstore-upload-url URL, the content of the response is
  * the URL that allows a user to upload a file to Blobstore. 
  * The user's browser uploads the file directly to the Blobstore via the generated URL.
- * Blobstore rewrites the request to contain a blobkey 
  */
 @WebServlet("/blobstore-upload-url")
 public class BlobstoreUploadUrlServlet extends HttpServlet {
@@ -41,8 +39,6 @@ public class BlobstoreUploadUrlServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String uploadUrl = blobstoreService.createUploadUrl("/image-upload-handler-blobstore");
-
-    // TODO: scan for inappropriate content: SafeSearchUtils.detectSafeSearch(uploadUrl);
 
     response.setContentType("text/html"); 
     response.getWriter().println(uploadUrl);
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/BlobstoreUploadUrlServlet.java
```diff
@@ -0,0 +1,50 @@
...
+ 
+  public BlobstoreUploadUrlServlet(){
+    blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+  }
+ 
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String uploadUrl = blobstoreService.createUploadUrl("/image-upload-handler-blobstore");
+
+    // TODO: scan for inappropriate content: SafeSearchUtils.detectSafeSearch(uploadUrl);
```
ofk-goog, line 45: I believe the place to implement this TODO is in the /image-upload-handler-blobstore servlet. This /blobstore-upload-url servlet doesn't have access to the image, which hasn't been uploaded yet.

src/main/java/com/google/sps/servlets/BlobstoreUploadUrlServlet.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.sps.utils.SafeSearchUtils;
+ 
+/**
+ * When the fetch() function requests the /blobstore-upload-url URL, the content of the response is
+ * the URL that allows a user to upload a file to Blobstore. 
+ * The user's browser uploads the file directly to the Blobstore via the generated URL.
+ * Blobstore rewrites the request to contain a blobkey
```
ofk-goog, line 30: Good documentation up until the last sentence, which doesn't seem relevant to this servlet.

src/main/java/com/google/sps/servlets/ServeImageServlet.java
```diff
@@ -0,0 +1,45 @@
...
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+ 
+/**
+ * Serves the blob by passing the blob key to blobstore service 
+ * replaces the getServingUrl method (Obtains a URL that can dynamically 
+ * serve the image stored as a blob)
```
ofk-goog, line 29: This is all great stuff to document. I recommend splitting it up, though: Here (above the class), document What the class accomplishes (e.g. "Serves the image file corresponding to the given blob key"); and within the class, document How it works and Why (e.g. "This is a workaround because the standard approach using getServingUrl() isn't working").

src/main/java/com/google/sps/servlets/BlobstoreUploadUrlServlet.java
EmmaWenzel: Alright, I moved the TODO there.

src/main/java/com/google/sps/servlets/BlobstoreUploadUrlServlet.java
EmmaWenzel: Got it, I took that out.

src/main/java/com/google/sps/servlets/ServeImageServlet.java
EmmaWenzel: That's split up, thanks!

PR #33, opened 2020-07-17
Create function getUserFromCookie

hannahsalameh:
Creates the function getUserFromCookie, which retrieves a user entity from the session ID cookie. It also fixes some bugs I created from not standardizing things between PRs.


src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -67,6 +66,32 @@ public static Entity getEntityFromDatastore(String entityName, String entityProp
...
+    * pair as one of its properties. This user entity is then returned. 
+    * 
+    * <p>
+    * 
+    * @param  cookies    an array of cookies (usually all cookies on front end)
+    * @param  datastore  the database where entities are stored
+    * @return            a single entity that has the cookie name/value pair as a property
+    */
+ 
+  public static Entity getUserFromCookie(Cookie cookies[], DatastoreService datastore){
```
ofk-goog, line 81: This is a great concept. It's exactly how you should be determining who the current user is. Good.

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -67,6 +66,32 @@ public static Entity getEntityFromDatastore(String entityName, String entityProp
...
+      if(datastore == null){
+          log.severe("Error in function getUserFromCookie(): Datastore passed in was null");
+          return null;
+      }
+    String sessionIDName = "SessionID";
+    Cookie cookie = CookieUtils.getCookieFromName(cookies, sessionIDName); //returns null if it doesn't exist
+    if(cookie == null){ 
+        log.severe("Error in function getUserFromCookie(): Cookie with name: " + sessionIDName + " not found.");
```
ofk-goog, line 89: I'd use `log.info` rather than `log.severe` for this. It's normal and expected for users sometimes to not be logged in, or even to have an invalid cookie.

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -67,6 +66,32 @@ public static Entity getEntityFromDatastore(String entityName, String entityProp
...
+     /**
+    * This function takes in a list of cookies and matches a sessionID cookie
+    * with a specific name/value pair to a user entity with the same name/value
+    * pair as one of its properties. This user entity is then returned. 
+    * 
+    * <p>
+    * 
+    * @param  cookies    an array of cookies (usually all cookies on front end)
+    * @param  datastore  the database where entities are stored
+    * @return            a single entity that has the cookie name/value pair as a property
```
ofk-goog, line 78: Is this the same "user entity" that's used elsewhere in the code?

If not, might it be better to return a type other than an entity? E.g. the userId, which can then be used to look up the user? Entities seems very hard to work with -- many of the changes I've been requesting on PRs lately have been related to handling of Entities.

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -67,6 +66,32 @@ public static Entity getEntityFromDatastore(String entityName, String entityProp
...
+    */
+ 
+  public static Entity getUserFromCookie(Cookie cookies[], DatastoreService datastore){
+      if(datastore == null){
+          log.severe("Error in function getUserFromCookie(): Datastore passed in was null");
+          return null;
+      }
+    String sessionIDName = "SessionID";
```
ofk-goog, line 86: If you make this a class constant (with `static final String sessionIdCookieName = "SessionID";` at the class level), then you can refer to that constant in your unit tests and other cookie-related code without the risk of misspelling "SessionID". (Which sounds silly but is the kind of thing that happens a lot and is hard to debug.)

src/main/java/com/google/sps/utils/UserUtils.java
hannahsalameh: this is the same user entity used elsewhere

hannahsalameh: u can get the user ID and all the user's info from the entity that is retrieved here

ofk-goog: Okay cool. Please mention in the documentation that this is the User entity. Currently the doc implies that it's something different

src/main/java/com/google/sps/utils/UserUtils.java
ofk-goog: Good, you did. Please also replace "SessionID" in the unit test file

src/main/java/com/google/sps/utils/UserUtils.java
ofk-goog: Actually nvm it's fine as is

src/test/java/com/google/sps/UserUtilTest.java
```diff
@@ -38,6 +38,7 @@
   // helper variable allows the use of entities in testing 
   private final LocalServiceTestHelper helper =
     new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
+  static final String sessionIdCookieName = "SessionID";
```
ofk-goog, line 41: Oh, you should use the `sessionIdCookieName` defined in UserUtil.java, rather than define a new one

PR #34, opened 2020-07-17
Make changes in game and user utils to only store one game per user

EmmaWenzel:
Gets rid of score entity assuming score will be stored in the user entity; stores only a single string for game id in user entity instead of an array list of strings; removes test to check if that array list is initialized, since it no longer needs to be.


src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -20,6 +20,7 @@
 import com.google.appengine.api.datastore.EntityNotFoundException;
 import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.datastore.Key;
+import com.google.sps.utils.UserUtils;
```
devonwasson, line 23: This adds UserUtils but GameUtils doesn't use UserUtils

EmmaWenzel: Yep, I took it out!

PR #37, opened 2020-07-20
gives user points for taking a quiz

colemanliyah:


src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -125,4 +125,24 @@ public String getNewQuestion(Entity game_entity, DatastoreService datastore) {
         return (update_game_entity.getProperty("quizQuestion")).toString();
     }
 
+    public Boolean giveUserPoints(Boolean userQuizStatus, Entity currentUser, DatastoreService datastore) {
```
EmmaWenzel, line 128: I think boolean should be lowercase here

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -125,4 +125,24 @@ public String getNewQuestion(Entity game_entity, DatastoreService datastore) {
...
+        if(currentUser == null || datastore == null) {
+            log.log(Level.SEVERE, "Given a null value in parameters");
```
EmmaWenzel, line 130: You don't need to do this if this method works, but it might be easier to do log.severe("Given a null value in parameters")

EmmaWenzel: You might also want to break this into two if statements so you can be more specific with the error messages

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
EmmaWenzel, line 128: maybe give a line or two of documentation for this function

colemanliyah: If I change it to lower case "b" I get an incompatible type error

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
colemanliyah: Okay I broke them up!

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
devonwasson: https://programming.guide/java/boolean-vs-boolean.html

We should use lowercase b here. Likely, the reason you're getting incompatibility errors is because if you only change it to lowercase b here but not everywhere else, the compiler will get angry.

devonwasson: It would help to maybe be a little more specific here as well. "Gives the user points" how many points? How does it give them to the user? What is a "user"?

Something like "Adds 20 points to the given User Entity for completing a quiz."

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -125,4 +125,29 @@ public String getNewQuestion(Entity game_entity, DatastoreService datastore) {
         return (update_game_entity.getProperty("quizQuestion")).toString();
     }
 
+    //Gives the user points if they have taken a quiz
+    public Boolean giveUserPoints(Boolean userQuizStatus, Entity currentUser, DatastoreService datastore) {
```
devonwasson, line 129: This function name is a bit ambiguous. If I writing code in this code base, I might think this function could be used to give an arbitrary number of points to a user. This is really a specific function for assigning points after taking a quiz and should be named as such to reflect that.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -125,4 +125,29 @@ public String getNewQuestion(Entity game_entity, DatastoreService datastore) {
...
+        }
+
+        if (datastore == null) {
+            log.log(Level.SEVERE, "Given a null {0}", datastore);
+            return null;
+        }
+
+        if(userQuizStatus) {
+            Key user_key = currentUser.getKey();
+            datastore.delete(user_key);
```
devonwasson, line 142: Deleting a user from the datastore and then adding a new one doesn't seem like the best idea here. What if something went wrong in between deleting and adding the user back to the database? We could lose all of their data. 

We should instead try and just update the values as needed.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -125,4 +125,29 @@ public String getNewQuestion(Entity game_entity, DatastoreService datastore) {
...
+
+            Entity updated_user = new Entity("user");
+            updated_user.setProperty("quiz_timestamp", System.currentTimeMillis());
+            updated_user.setProperty("userID", currentUser.getProperty("userID"));
+            updated_user.setProperty("score", ((int) currentUser.getProperty("score")) + 20);
+            datastore.put(updated_user);
```
devonwasson, line 148: This could be generalized and very useful later. If we pull the logic of adding points to a user entity into a helper function (maybe UserUtils) then we can use this in multiple places. For example, we'll want to give points to a user if they are picked in a quiz, so we'll need to use a generalized version of this logic there as well.

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -190,4 +190,32 @@ public void getNewQuestion_validParamters() {
             Assert.assertEquals(true, true);
         }
     }
+
+     @Test 
+    //Test if giveUserPoints is given null value(s) as a parameter
+    public void giveUserPoints_nullParameters() {
```
devonwasson, line 196: this test never tests a null datastore. We should have one test for null entity and one for null datastore.

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -190,4 +190,32 @@ public void getNewQuestion_validParamters() {
...
+        QuizTimingPropertiesUtils timing_properties_test = new QuizTimingPropertiesUtils();
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        
+        Entity user = new Entity("user");
+        user.setProperty("quiz_timestamp", 159430944365L);
+        user.setProperty("userID", 12345);
+        user.setProperty("score", 0);
+        datastore.put(user);
+ 
+        Boolean actual = timing_properties_test.giveUserPoints(true, user, datastore);
```
devonwasson, line 217: what about a test where userQuizStatus is false?

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
colemanliyah: I think the only thing that would be replicated here is updating the score because the timestamp only needs to be updated this once when involving the user taking the quiz. Also, I just reduced all the code around it

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -190,21 +190,43 @@ public void getNewQuestion_validParamters() {
...
  
-        Boolean actual = timing_properties_test.giveUserPoints(true, user, datastore);
-        Assert.assertEquals(null, actual);
+        boolean actual = timing_properties_test.giveUserQuizTakenPoints(true, user, datastore);
+        Assert.assertEquals(false, actual);
+    }
+ 
+    @Test 
+    //Test if giveUserPoints is given null datastore value as a parameter
+    public void giveUserQuizTakenPoints_nullParameters() {
```
devonwasson, line 207: this should be changed to nullDatastore

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -214,8 +236,7 @@ public void giveUserPoints_validParameters() {
...
         datastore.put(user);
  
-        Boolean actual = timing_properties_test.giveUserPoints(true, user, datastore);
+        boolean actual = timing_properties_test.giveUserQuizTakenPoints(true, user, datastore);
         Assert.assertEquals(true, actual);
     }
-
-}
\ No newline at end of file
+}
```
devonwasson: We should add another test that makes sure that a user's score is actually increased. So one where we make a user entity (maybe with a non-zero and a zero score, so two test cases), then add 20 points, and ensure their score is score + 20.

Also, we should add a test case where the user never had a score property set (aka this line never exists in the test: user.setProperty("score", 0);)

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -107,22 +119,42 @@ public Boolean isQuizOutdated(Long current_quiz_time) {
...
+            log.log(Level.SEVERE, "Given a null {0}", datastore);
+            return false;
+        }
+
+        if(userQuizStatus) {
+            try {
+                currentUser.setProperty("score", ((Number) currentUser.getProperty("score")).intValue() + 20);
+            } catch (NullPointerException e) {
+                log.log(Level.SEVERE, "User Entity has no 'score' property");
+                return false;
```
devonwasson, line 152: If user has no score property, maybe we can be clever and assign them one here? I know the user *should* already have it, but if they don't it would make life easier if we create the property with a default value of whatever the points score is.

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -46,10 +46,10 @@ public void tearDown() {
     //Test where user has null timestamp
     public void getTimestampProperty_nullUserTimeStamp() {
         QuizTimingPropertiesUtils timing_properties_test = new QuizTimingPropertiesUtils();
-
+
```
devonwasson, line 49: I think you added a bunch of spaces to this file for some reason. Removing them will make the PR much more readable and follow formatting guides anyway.

src/test/java/com/google/sps/TimingPropertiesTest.java
devonwasson: please change the function name.

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -190,4 +190,108 @@ public void getNewQuestion_validParamters() {
...
+        user.setProperty("userID", 12345);
+        int score_value = 15;
+ 
+        user.setProperty("score", score_value);
+        datastore.put(user);
+ 
+        boolean actual = timing_properties_test.giveUserQuizTakenPoints(true, user, datastore);
+        if(((int) user.getProperty("score")) == (score_value + 20) ) {
+            Assert.assertEquals(actual, true);
+        }
```
devonwasson, line 259: So in this case, we actually don't want to test the result of giveUserQuizTakenPoints(). We want to assert on assertEquals(entity.getProperty("score"), 35);

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
devonwasson: yes, a good reason to refactor out updating the score is because what if we want to assign more or less than 20 points for taking a quiz? If we have the number 20 hard coded as a part of this method, then our test cases will all break (as they rely on the number 20 being the exact number). Instead, if we take this method and refactor it out and have it take a generic number of points, we can test that method and never have to worry about the tests breaking. Then, when we test the larger method, we don't need to test that the user's score actually increased because we can assume that calling that function is an already well-tested method.

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -190,4 +190,108 @@ public void getNewQuestion_validParamters() {
...
+        user.setProperty("quiz_timestamp", 159430944365L);
+        user.setProperty("userID", 12345);
+        int score_value = 0;
+        user.setProperty("score", score_value);
+        datastore.put(user);
+ 
+        boolean actual = timing_properties_test.giveUserQuizTakenPoints(true, user, datastore);
+        if(((int) user.getProperty("score")) == (score_value + 20) ) {
+            Assert.assertEquals(actual, true);
+        }
```
devonwasson, line 278: same as the comment above re: asserting on the actual score.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -22,6 +22,14 @@
 import java.util.List;
 import java.util.ArrayList;
 import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Key;
+import java.text.DateFormat;
+import java.util.Date;
+import java.util.Random;
```
devonwasson, line 31: random is a duplicate import

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -22,6 +22,14 @@
 import java.util.List;
 import java.util.ArrayList;
 import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Key;
+import java.text.DateFormat;
+import java.util.Date;
```
devonwasson, line 30: are any of these imports being used?

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -22,6 +22,14 @@
...
+import java.util.Random;
+import java.util.logging.Logger;
```
devonwasson, line 32: Logger is a duplicate import.

src/test/java/com/google/sps/TimingPropertiesTest.java
colemanliyah: Is there a general rule of thumb for spaces that adheres to the formatting guide lines?

PR #38, opened 2020-07-20
Game page html and js file

colemanliyah:
The JS file calls on the servlet to check if the user has taken a quiz or not yet. If no, then the "take a quiz button" will show up on the gameBoard.html page. Otherwise, the button to take a quiz will be hidden

### DIFF
--- a/src/main/webapp/gameFunctions.js
+++ b/src/main/webapp/gameFunctions.js
@@ -1,0 +1,7 @@
+function getUserQuizStatus() {
+    fetch("/user-quiz-status-servlet").then(response => response.json()).then((tasks) => {
+        if(tasks==false){
+            document.getElementById("QuizButton").style.display = "block";
+        }
+    });
+}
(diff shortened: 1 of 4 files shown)


src/main/webapp/gameFunctions.js
```diff
@@ -0,0 +1,7 @@
+function getUserQuizStatus() {
+    fetch("/user-quiz-status-servlet").then(response => response.json()).then((tasks) => {
```
ofk-goog, line 2: Instead of `tasks`, how about a descriptive name like `user_has_taken_quiz`?

src/main/webapp/gameFunctions.js
```diff
@@ -0,0 +1,7 @@
...
+        if(tasks==false){
```
ofk-goog, line 3: Nit: Instead of comparing `x == false`, do `!x`. This is mostly for the sake of style.

PR #40, opened 2020-07-20
Add start game servlet to create game entity

EmmaWenzel:
Start game servlet creates a game if a user doesn't already have a game, and adds the user to the game and the game to the user

### DIFF
--- a/src/main/java/com/google/sps/servlets/startGameServlet.java
+++ b/src/main/java/com/google/sps/servlets/startGameServlet.java
@@ -45,31 +45,37 @@ public class startGameServlet extends HttpServlet {
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
  
     String gameName = request.getParameter("game-name");
-    
-    // query user 
-    Cookie cookies[] = request.getCookies();
-    Entity userEntity = UserUtils.getUserFromCookie(cookies, datastore);
  
-    String newGame = GameUtils.joinGame(userEntity, gameName, datastore, null);
- 
-    if(newGame == "noGameName"){
+    if(gameName == ""){
       response.sendRedirect("/start.html");
       return;
     }
+    
+    // query user 
+    Cookie cookies[] = request.getCookies();
+    if(cookies == null){
+        response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+        return;
+    }
+    Entity userEntity = UserUtils.getUserFromCookie(cookies, datastore);
+    if(userEntity == null){
+        response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+        return;
+    }
  
-    if(newGame == "nullUserEntity" || newGame == "createGameFailed" || newGame == "userAddedFailed" || newGame == "gameAddedFailed"){
+    // prevent users from joining more than one game
+    if(userEntity.getProperty("gameId") != null){
+      response.sendRedirect("/gameBoard.html");
+      return;
+    }
+ 
+    Entity newGame = GameUtils.startGame(gameName, datastore);
+    boolean gameSet = GameUtils.setGame(userEntity, datastore, newGame);
+ 
+    if(newGame == false){
       response.sendRedirect("/index.html");
       return;
-    }
- 
-    if(newGame == "alreadyHasGame"){
-      PrintWriter out = response.getWriter();
-      out.println("<p>You are already in a game, see the gameboard here:</p>");
-      out.println("<a href = \"gameBoard.html\"><h3>Your game</h3></a>");
-      return;
-    }
- 
-    if(newGame == "success"){
+    }else{
       response.sendRedirect("/gameBoard.html");
       return;
     }
--- a/src/main/java/com/google/sps/servlets/startGameServlet.java
+++ b/src/main/java/com/google/sps/servlets/startGameServlet.java
@@ -72,7 +72,7 @@ public class startGameServlet extends HttpServlet {
     Entity newGame = GameUtils.createGameEntity(gameName, datastore);
     boolean gameSet = GameUtils.setGame(userEntity, datastore, newGame);
  
-    if(newGame == false){
+    if(gameSet == false){
       response.sendRedirect("/index.html");
       return;
     }else{
--- a/src/main/java/com/google/sps/servlets/startGameServlet.java
+++ b/src/main/java/com/google/sps/servlets/startGameServlet.java
@@ -69,7 +69,7 @@ public class startGameServlet extends HttpServlet {
       return;
     }
  
-    Entity newGame = GameUtils.startGame(gameName, datastore);
+    Entity newGame = GameUtils.createGameEntity(gameName, datastore);
     boolean gameSet = GameUtils.setGame(userEntity, datastore, newGame);
  
     if(newGame == false){
--- a/src/main/java/com/google/sps/servlets/startGameServlet.java
+++ b/src/main/java/com/google/sps/servlets/startGameServlet.java
@@ -1,0 +1,85 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+ 
+package com.google.sps.servlets;
+ 
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.sps.utils.GameUtils;
+import com.google.sps.utils.UserUtils;
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
+import java.io.PrintWriter;
+ 
+/** Create game entity and connect it to the user, then update the user entity to include the game id
+    and store both in datastore when the user creates a new game.
+ */
+@WebServlet("/start-game")
+public class startGameServlet extends HttpServlet {
+ 
+  DatastoreService datastore;
+ 
+  public startGameServlet(){
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+ 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+ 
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.servlet.http.*;
+import java.io.PrintWriter;
+ 
+/** Create game entity and connect it to the user, then update the user entity to include the game id
+    and store both in datastore when the user creates a new game.
+ */
+@WebServlet("/start-game")
+public class startGameServlet extends HttpServlet {
+ 
+  DatastoreService datastore;
```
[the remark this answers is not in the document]
EmmaWenzel: I can't declare it within the constructor (because then it can't be used in the function below), but having it initialized in the constructor makes it so it only has to be called once

src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -0,0 +1,85 @@
...
+ 
+  public startGameServlet(){
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+ 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+ 
+    String gameName = request.getParameter("game-name");
```
devonwasson, line 47: Do we want to check for empty string game names?

src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -0,0 +1,85 @@
...
+    Entity userEntity = UserUtils.getUserFromCookie(cookies, datastore);
+ 
+    if(userEntity == null){
+        // no user entity, send back to login
+        response.sendRedirect("/index.html");
+        return;
+    }
+ 
+    // prevent users from joining more than one game
+    if((String) userEntity.getProperty("gameId") != null){
```
devonwasson, line 60: We shouldn't need to cast this to a string if we're checking against null

src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -0,0 +1,85 @@
...
+        userAdded = GameUtils.addUserToGame((String) userEntity.getProperty("userId"), gameEntity, datastore);
+        gameAdded = UserUtils.addGameToUser(userEntity, datastore, (String) gameEntity.getProperty("gameId"));
+    }
+    
+    if(userAdded && gameAdded){
+        response.sendRedirect("/gameBoard.html");
+    }else{
+        // something's wrong, send back to login
+        response.sendRedirect("/index.html");
+    }
```
devonwasson, line 83: This is a pretty vague else clause. We have no idea what had to happen for us to get here. Is userAdded False? Is gameAdded False? What does it mean if either of those are false?

This would be a good place to do some logging. I would suggest a little rework of the flow from getting the gameEntity to the end of file. First, just return early if gameEntity is null rather than checking to see if it's not null. Then, check the status of addUserToGame(). If it's False, log why then redirect to index.html. Then do the same for addGameToUser. At this point, we've checked all there is to check so we can safely just redirect to gameBoard.html without needed to do any more checks.

src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -0,0 +1,85 @@
...
+public class startGameServlet extends HttpServlet {
+ 
+  DatastoreService datastore;
+ 
+  public startGameServlet(){
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+ 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
devonwasson, line 45: This is a lot of code so we should test this. It would be nice to test using a framework that allows us to test servlets but considering we've used the helper class pattern to test complex code before, it might be best to add the meat of this logic to a helper function and test that function. That can be done in this same PR. Just add it to an existing utility class (GameUtility?)

src/main/java/com/google/sps/servlets/startGameServlet.java
EmmaWenzel: Yes, I added a check to the helper function

src/main/java/com/google/sps/servlets/startGameServlet.java
EmmaWenzel: Alright, I took (String) out

src/main/java/com/google/sps/servlets/startGameServlet.java
EmmaWenzel: This is all broken up in the helper function

src/main/java/com/google/sps/servlets/startGameServlet.java
EmmaWenzel: Alright, that's done and tested

src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -123,4 +124,59 @@ public static boolean addUserToGame(String userId, Entity gameEntity, DatastoreS
...
     return true;
   }
-}
\ No newline at end of file
+ 
+  public static String joinGame(Entity userEntity, String gameName, DatastoreService datastore, String gameId) {
+ 
+    if(gameName == ""){
+        return "noGameName";
+    }
```
ofk-goog, line 132: There's no need for this check. Currently it's already impossible to create a game with an empty name. But if you ever change your mind and allow empty games, you'll need to delete this check. So this check only hurts you.

src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -123,4 +124,59 @@ public static boolean addUserToGame(String userId, Entity gameEntity, DatastoreS
     
     return true;
   }
-}
\ No newline at end of file
+ 
+  public static String joinGame(Entity userEntity, String gameName, DatastoreService datastore, String gameId) {
```
ofk-goog, line 128: This pattern -- returning a String to indicate the outcome of a function call -- is error-prone and makes it difficult to use `joinGame()` properly. It's better practice to assume the caller will test for null, and throw `NullPointerExeption` if they do not.

As for the case where the user is already in the game, there are two ways to handle it:
1. Throw a custom `Exception` subclass, called e.g. `UserAlreadyInGameException`.
2. Don't tell the caller whether you added the user, or whether the user was already in the game. Does the caller really need to know?

I prefer option 2, because it makes life simpler for the caller. If they really needed to know whether the user was part of the game already, they could have called `IsValidGameName`.

src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -50,36 +50,28 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
     }
  
-    // prevent users from joining more than one game
-    if((String) userEntity.getProperty("gameId") != null){
+    if(newGame == "alreadyHasGame"){
       PrintWriter out = response.getWriter();
       out.println("<p>You are already in a game, see the gameboard here:</p>");
       out.println("<a href = \"gameBoard.html\"><h3>Your game</h3></a>");
       return;
     }
```
devonwasson: We could probably just skip this and redirect them right away to the gameboard.

src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -11,21 +11,22 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
+
```
devonwasson, line 14: There are a lot of lines with a single space added in this PR. If those are removed, it would make the PR much more readable and follow the style guides anyway.

src/main/java/com/google/sps/utils/GameUtils.java
devonwasson: I suggested this check because you'll never want to create a game with an empty name (from a usability standpoint) so it's not bad to make sure it's not happening. 

Why do you think it's impossible to create a game with an empty name?

src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -123,4 +124,59 @@ public static boolean addUserToGame(String userId, Entity gameEntity, DatastoreS
...
+    
+        if(gameEntity == null){ 
+          return "badGameCode";
+        }
+    }
+ 
+    // add user to game entity + vice versa
+    boolean userAdded = GameUtils.addUserToGame((String) userEntity.getProperty("userId"), gameEntity, datastore);
+    if(!userAdded){
+        log.severe("failed to add user to game " + gameName);
```
devonwasson, line 170: Which user? could be useful to log the user ID

src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -123,4 +124,59 @@ public static boolean addUserToGame(String userId, Entity gameEntity, DatastoreS
...
+        return "userAddedFailed";
+    }
+    boolean gameAdded = UserUtils.addGameToUser(userEntity, datastore, (String) gameEntity.getProperty("gameId"));
+    if(!gameAdded){
+        log.severe("failed to add game "  + gameName + " to user");
```
devonwasson, line 175: same comment as above (helps to specify the user)

src/main/java/com/google/sps/utils/GameUtils.java
devonwasson: +1

src/main/java/com/google/sps/utils/GameUtils.java
ofk-goog: I assumed `joinGame` was a function for joining an existing game. Does it also create a game? If so it probably shouldn't

EmmaWenzel: Sounds good, I removed this check.

src/main/java/com/google/sps/utils/GameUtils.java
EmmaWenzel: Alright, I changed the function and moved some of the checks around; some of it's back in the servlet, and the function only returns true or false. It can still be called by both start and join, because it is called after the game is created.

src/main/java/com/google/sps/servlets/startGameServlet.java
EmmaWenzel: That makes a lot more sense; that's changed.

src/main/java/com/google/sps/utils/GameUtils.java
EmmaWenzel: I think in all of these places a line was both added and removed (so like net 0 line); I think it was probably caused by something shifting?

src/main/java/com/google/sps/utils/GameUtils.java
EmmaWenzel: I added the user to these.

src/main/java/com/google/sps/utils/GameUtils.java
EmmaWenzel: Is there a way to change the way this looks in the pr?

ofk-goog: You added a space to an otherwise empty line. As in the ` ` character you get when you press the spacebar.

EmmaWenzel: Oh shoot I totally missed that, yep. Those are taken out

PR #42, opened 2020-07-20
Add servlets to store and retrieve blobkeys to/from blobstore

EmmaWenzel:
UploadHandlerServlet is revised to use helper functions and get the correct user entity; getBlobKeyServlet gets and returns the stored blobkey so that it can be used to create a url for the image

### DIFF
--- a/src/main/java/com/google/sps/servlets/UploadHandlerServlet.java
+++ b/src/main/java/com/google/sps/servlets/UploadHandlerServlet.java
@@ -1,0 +1,69 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+ 
+package com.google.sps.servlets;
+ 
+import com.google.appengine.api.blobstore.BlobKey;
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.io.IOException;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.sps.utils.UserUtils;
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*; 
+ 
+/** Blobstore upload handler: gets blobkey (Blobstore rewrites the request to contain a blobkey) 
+    for an uploaded image and stores it in datastore 
+    (see https://cloud.google.com/appengine/docs/standard/java/blobstore#3_implement_upload_handler
+    for documentation of an upload handler)
+ */
+@WebServlet("/image-upload-handler-blobstore")
+public class UploadHandlerServlet extends HttpServlet {
+ 
+  BlobstoreService blobstoreService;
+  DatastoreService datastore;
+ 
+  public UploadHandlerServlet(){
+    blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+ 
+  // TODO: Limit this so it can only be used as a blobstore upload handler; cut off end-user access
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    
+    // TODO: userID using new func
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/UploadHandlerServlet.java
```diff
@@ -0,0 +1,69 @@
...
+  public UploadHandlerServlet(){
+    blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+    datastore = DatastoreServiceFactory.getDatastoreService();
+  }
+ 
+  // TODO: Limit this so it can only be used as a blobstore upload handler; cut off end-user access
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    
+    // TODO: userID using new func
```
ofk-goog, line 56: This TODO appears to be done

src/main/java/com/google/sps/servlets/UploadHandlerServlet.java
```diff
@@ -0,0 +1,69 @@
...
+    Cookie cookies[] = request.getCookies();
+    Entity userEntity = UserUtils.getUserFromCookie(cookies, datastore);
```
ofk-goog, line 58: Remember to handle the case where this returns `null` (because of an invalid/missing cookie)

src/main/java/com/google/sps/servlets/getBlobKeyServlet.java
```diff
@@ -0,0 +1,66 @@
...
+    Entity userEntity = UserUtils.getUserFromCookie(cookies, datastore);
+ 
+    String blobKey = "noKey";
+    if(userEntity != null){
+      blobKey = (String) userEntity.getProperty("blobKey");
+    }
+    if(blobKey == null){
+        // if the user hasn't uploaded a picture yet, nothing is printed
+        blobKey = "noKey";
+    }
```
ofk-goog, line 60: This is a good opportunity to send an HTTP error response instead of completing the request as normal. Liyah learned how to do this a few weeks ago, so ask her. I forget how.

Pick the most applicable code from this list: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes -- it's a fun read IMO

src/main/java/com/google/sps/servlets/getBlobKeyServlet.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.sps.utils.UserUtils; 
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
+ 
+/** Retrieves blobkey from Datastore */
```
ofk-goog, line 36: More specific plz. It gets the blobKey for the image belonging to the logged-in user, right?

src/main/java/com/google/sps/servlets/UploadHandlerServlet.java
EmmaWenzel: I missed that one, thanks!

src/main/java/com/google/sps/servlets/UploadHandlerServlet.java
EmmaWenzel: Got it, I used the http error response here.

src/main/java/com/google/sps/servlets/getBlobKeyServlet.java
EmmaWenzel: Yes, I changed that comment.

src/main/java/com/google/sps/servlets/getBlobKeyServlet.java
EmmaWenzel: The way I have this error set up is connected to the javascript function where it's called, instead of getting errors when there isn't a picture saved, it just doesn't print a picture. Because this function can easily be called if the user doesn't have a picture, this seemed better; I can make a function to hide the button but that might be a stretch goal

src/main/java/com/google/sps/servlets/getBlobKeyServlet.java
```diff
@@ -46,17 +46,21 @@ public getBlobKeyServlet(){
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
  
-    // get user id -- get entity from user id -- new func!!
+    // get the user entity
     Cookie cookies[] = request.getCookies();
+    if(cookies == null){
+        response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
```
ofk-goog, line 52: Nice job. You should also do this if `userEntity` is null.

src/main/java/com/google/sps/servlets/getBlobKeyServlet.java
ofk-goog: A response code is a more robust way to signal the error to the javascript, regardless of how the javascript will handle the error. Please use a response code.

src/main/java/com/google/sps/servlets/getBlobKeyServlet.java
EmmaWenzel: Okay, done.

src/main/java/com/google/sps/servlets/getBlobKeyServlet.java
```diff
@@ -46,17 +46,23 @@ public getBlobKeyServlet(){
...
+        return;
     }
+ 
+    String blobKey = (String) userEntity.getProperty("blobKey");
     if(blobKey == null){
-        // if the user hasn't uploaded a picture yet, nothing is printed
-        blobKey = "noKey";
+      // if the user hasn't uploaded a picture yet, nothing is printed
+      response.setStatus(HttpServletResponse.SC_NOT_FOUND);
```
ofk-goog, line 64: Good choice of error code

PR #47, opened 2020-07-21
add function to add points to a user and test the function

EmmaWenzel:
This function adds a specified number of points to the user; the tests cover if the user already has points and if the user doesn't already have points. This also fixes a typo in the sessionIDcookieName variable

### DIFF
--- a/src/main/java/com/google/sps/utils/UserUtils.java
+++ b/src/main/java/com/google/sps/utils/UserUtils.java
@@ -30,7 +30,7 @@ import com.google.appengine.api.datastore.Query.FilterPredicate;
 import java.util.logging.Logger;
 
 public final class UserUtils {
-    static final String sessionIdCookieName = "SessionID";
+    public static final String sessionIDCookieName = "SessionID";
     private static final Logger log = Logger.getLogger(UserUtils.class.getName());
 
     /**
@@ -137,5 +137,17 @@ public final class UserUtils {
     datastore.put(userEntity);
  
     return true; 
+  }
+
+  /**
+    adds a specified number of points to the user's points
+  */
+  public static void addPoints(Entity userEntity, int numPoints, DatastoreService datastore){
+    try {
+      userEntity.setProperty("score", ((int) userEntity.getProperty("score")) + numPoints);
+    } catch (NullPointerException e) {
+      userEntity.setProperty("score", numPoints);                
+    }
+    datastore.put(userEntity);
   }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -138,4 +138,16 @@ public static boolean addBlobKey(String blobKey, Entity userEntity, DatastoreSer
  
     return true; 
   }
+
+  /**
+    adds a specified number of points to the user's points
+  */
+  public static void addPoints(Entity userEntity, int numPoints, DatastoreService datastore){
+    try {
+      userEntity.setProperty("score", ((int) userEntity.getProperty("score")) + numPoints);
```
colemanliyah, line 147: Instead of casting it as int i would cast it as a (Number) so that way you don't have to cast it as a long down below when you write your assert statements

src/test/java/com/google/sps/UserUtilTest.java
```diff
@@ -386,4 +386,35 @@ public void noCookies(){
...
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Entity user1 = new Entity("user");
+    user1.setProperty("userId", "newUser1");
+    datastore.put(user1);
+
+    UserUtils.addPoints(user1, 20, datastore);
+
+    Entity userEntity = UserUtils.getEntityFromDatastore("user", "userId", "newUser1", datastore);
+    
+    Assert.assertEquals(20, (long) userEntity.getProperty("score"));
```
colemanliyah, line 418: I think you should change this so that you don't have to cast a long value here, my comment above gives a suggestion on how you might be able to do this

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: Okay, that's done

src/test/java/com/google/sps/UserUtilTest.java
EmmaWenzel: I still need to cast the "expected" to long to avoid errors, but otherwise it works

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -30,7 +30,7 @@
 import java.util.logging.Logger;
 
 public final class UserUtils {
-    static final String sessionIdCookieName = "SessionID";
+    public static final String sessionIDCookieName = "SessionID";
```
ofk-goog, line 33: As long as you're changing this name, please follow the Google Java Style Guide and rename it to `SESSION_ID_COOKIE_NAME`. (Hideous, I know.)

See https://google.github.io/styleguide/javaguide.html#s5-naming

PR #48, opened 2020-07-22
Renaming and Rewriting Functions 

colemanliyah:
I changed the parameters accepted by on the functions and adjusted their respective tests. I also did some renaming

### DIFF
--- a/src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
+++ b/src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
@@ -27,11 +27,11 @@ import java.util.logging.Logger;
 import java.util.logging.Level;
 import java.util.TimeZone;
 import java.text.SimpleDateFormat;
-
+ 
 public final class QuizTimingPropertiesUtils {
-
+ 
     private static final Logger log = Logger.getLogger(QuizTimingPropertiesUtils.class.getName());
-
+ 
     List<String> quiz_questions = new ArrayList<String>(List.of(
         "Which plant has the most food growing from it?",
         "Which plant has the prettiest colors?",
@@ -52,9 +52,9 @@ public final class QuizTimingPropertiesUtils {
         "Which plant would look the best outside in a garden?",
         "Which plant would you give as a gift?"
     ));
-
+ 
     //This function gets the the "quiz_timestamp" property of the entity that is fed into the function
-    public Long getTimestampProperty(String entity, DatastoreService datastore) {
+    public Long getQuizTimestampProperty(String entity, String idProperty, String idPropertyValue, DatastoreService datastore) {
         Query query = new Query(entity);
         PreparedQuery pq;
         try {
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -44,63 +44,67 @@ public void tearDown() {
...
-    public void getTimestampProperty_noDatastore() {
+    public void getQuizTimestampProperty_noDatastore() {
         QuizTimingPropertiesUtils timing_properties_test = new QuizTimingPropertiesUtils();
  
         Entity user = new Entity("user");
-        Long actual = timing_properties_test.getTimestampProperty("user", null);
+        user.setProperty("userID", "abc");
+ 
+        Long actual = timing_properties_test.getQuizTimestampProperty("user", "userID", (user.getProperty("userID")).toString(), null);
         Assert.assertEquals(null, actual);
```
EmmaWenzel, line 79: This might be one of those where you want to check for a null pointer exception instead of returning null

colemanliyah: so I have try catch statements in the helper functions which is why I don't have them in the test. That is why the tests are still returning null. I can move the try catch statements over but I just thought it would be cleaner to keep it in the helper functions

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -29,9 +29,9 @@
 import java.text.SimpleDateFormat;
 
 public final class QuizTimingPropertiesUtils {
-
+
```
devonwasson, line 32: A bunch of space characters were added to this file. Removing those will make the number of "changed" lines go down and make this PR easier to review. Also, we shouldn't be adding random spaces to empty lines anyway.

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -52,9 +52,9 @@
         "Which plant would look the best outside in a garden?",
         "Which plant would you give as a gift?"
     ));
-
+ 
     //This function gets the the "quiz_timestamp" property of the entity that is fed into the function
-    public Long getTimestampProperty(String entity, DatastoreService datastore) {
+    public Long getQuizTimestampProperty(String entity, String idProperty, String idPropertyValue, DatastoreService datastore) {
```
devonwasson, line 57: idProperty and idPropertyValue are ambiguous. What do they actually represent? the variable names should be altered to reflect that.

src/test/java/com/google/sps/TimingPropertiesTest.java
devonwasson: Emma is right, it would be fine to not try/catch at all and just let the NPE happen (and test to make sure it happens) but we can save that for a future PR dedicated to cleaning up things like this.

PR #50, opened 2020-07-22
GameQuizStatusServlet

colemanliyah:
The servlet gets the quiz question. It will get a new question if the old one is outdated.

### DIFF
--- a/src/main/java/com/google/sps/servlets/GameQuizStatusServlet.java
+++ b/src/main/java/com/google/sps/servlets/GameQuizStatusServlet.java
@@ -1,0 +1,52 @@
+package com.google.sps.servlets;
+ 
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Key;
+import com.google.sps.QuizTimingPropertiesUtils;
+import com.google.sps.UserUtils;
+ 
+//This servlet gets the quiz question from the Game entity 
+//The quiz question is updated if the quiz timestamp is outdated
+@WebServlet("/game-quiz-status-servlet")
+public class GameQuizStatusServlet extends HttpServlet {
+ 
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+ 
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        Cookie cookies[] = request.getCookies();
+        if(cookies == null){
+            response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            return;
+        }
+        Entity userEntity = UserUtils.getUserFromCookie(cookies, datastore);
+        if(userEntity == null){
+            response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            return;
+        }
+ 
+        QuizTimingPropertiesUtils timing_properties = new QuizTimingPropertiesUtils();
+        UserUtils user_utils_class = new UserUtils();
+        Long current_quiz_stamp = timing_properties.getQuizTimestampProperty("Game", "currentGame", userEntity.getProperty("currentGame").toString(), datastore);
+        Entity current_game = user_utils_class.getEntityFromDatastore("Game", "currentGame", userEntity.getProperty("currentGame").toString(), datastore);
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/GameQuizStatusServlet.java
```diff
@@ -0,0 +1,52 @@
...
+            return;
+        }
+        Entity userEntity = UserUtils.getUserFromCookie(cookies, datastore);
+        if(userEntity == null){
+            response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            return;
+        }
+ 
+        QuizTimingPropertiesUtils timing_properties = new QuizTimingPropertiesUtils();
+        UserUtils user_utils_class = new UserUtils();
```
hannahsalameh, line 39: u could just make this a static function so you wouldn't need a class instance

colemanliyah: you mean timing_properties? It is actually the whole class that I am instantiating here

ofk-goog: Instead of instantiating these classes, make the methods you need static. Then you can call `UserUtils.getEntityFromDatastore("Game", "currentGame", etc, etc)` without making an instance of `UserUtils`. And ditto for `QuizTimingPropertyUtils`.

src/main/java/com/google/sps/servlets/GameQuizStatusServlet.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Key;
+import com.google.sps.QuizTimingPropertiesUtils;
+import com.google.sps.UserUtils;
+ 
+//This servlet gets the quiz question from the Game entity 
+//The quiz question is updated if the quiz timestamp is outdated
```
ofk-goog, line 20: Please rewrite this for a reader who doesn't know what a Game entity is. Also, please mention what data is required in the request, and what data will be in the response.

colemanliyah: I am not sure what to write for the data for the reponse and request because I do not use these parameters in this servlet

src/main/java/com/google/sps/servlets/GameQuizStatusServlet.java
colemanliyah: Did this!

src/main/java/com/google/sps/servlets/GameQuizStatusServlet.java
```diff
@@ -0,0 +1,52 @@
...
+        Entity current_game = user_utils_class.getEntityFromDatastore("Game", "currentGame", userEntity.getProperty("currentGame").toString(), datastore);
+
+        Gson gson = new Gson();
+        response.setContentType("application/json;");
+ 
+        if(timing_properties.isTimestampOutdated(current_quiz_stamp)) {
+            response.getWriter().println(gson.toJson(timing_properties.getNewQuestion(current_game, datastore)));
+        }
+        
+        response.getWriter().println(gson.toJson(current_game.getProperty("quizQuestion")));
```
ofk-goog: If the quiz is outdated, the response will contain two JSON objects separated by a newline. That seems like a bug.

src/main/java/com/google/sps/servlets/GameQuizStatusServlet.java
ofk-goog: You use a cookie from the request to determine the current logged-in user. And you write JSON data to the response. It'd be good to document what that JSON data looks like.

Also, the "gets the quiz question from the Game entity" framing only makes sense to you, Hannah, and Emma (at most). Try pretending the reader has used your finished website but hasn't looked at the code before. How would you explain this servlet to that reader?

src/main/java/com/google/sps/servlets/GameQuizStatusServlet.java
colemanliyah: If the quiz is outdated, it returns the response in he if statement and then it will not reach the statement on line 50

src/main/java/com/google/sps/servlets/GameQuizStatusServlet.java
colemanliyah: Got it!

src/main/java/com/google/sps/servlets/GameQuizStatusServlet.java
ofk-goog: I don't see a `return` statement. I just see `response.getWriter().println(...)`

src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -34,6 +34,7 @@
     private static final Logger log = Logger.getLogger(QuizTimingPropertiesUtils.class.getName());
 
     static final List<String> quiz_questions = new ArrayList<String>(List.of(
+        "Which plant has the most food growing from it?",
```
ofk-goog, line 37: I'm not sure how this happened. But this line is now a duplicate

PR #51, opened 2020-07-22
Fix property sets and gets to all use the same names, fix a few compi…

EmmaWenzel:
…le time errors. Remove gameScores property, isValidGameName function, and isValidGameName tests (outdated because a user can only have one game). Fix syntax errors in joinGame. Change instances of "userId" to the correct "userID". Fix syntax of list declaration in QuizTimingPropertiesUtils.

### DIFF
--- a/src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
+++ b/src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
@@ -32,27 +32,26 @@ public final class QuizTimingPropertiesUtils {
 
     private static final Logger log = Logger.getLogger(QuizTimingPropertiesUtils.class.getName());
 
-    List<String> quiz_questions = new ArrayList<String>(List.of(
-        "Which plant has the most food growing from it?",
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/sps/utils/QuizTimingPropertiesUtils.java
```diff
@@ -32,27 +32,26 @@
...
+        add("Which plant looks the most serene?");
+        add("Which plant needs the most sunlight?");
+        add("Which plant needs the most water?");
+        add("Which plant represents you?");
+        add("Which plant represents your best friend?");
+        add("Which plant will most likely impress your friends and family?");
+        add("Which plant would look the best inside as a houseplant?");
+        add("Which plant would look the best outside in a garden?");
+        add("Which plant would you give as a gift?");
+    }};
```
devonwasson, line 54: Does doing something like this work?

List<String> supplierNames = Arrays.asList("string1", "string2", "string3");

Would make this code simpler.

PR #52, opened 2020-07-23
Upload points function

EmmaWenzel:
Add upload points function checks if it has been more than a day since the user uploaded using the isTimestampOutdated function; if it has been more than a day or the user has never uploaded, it adds 20 points to the user's score. The tests test for if the user has never uploaded, if the user has uploaded less than a day ago, and if the user has uploaded more than a day ago.

### DIFF
--- a/src/main/java/com/google/sps/utils/UserUtils.java
+++ b/src/main/java/com/google/sps/utils/UserUtils.java
@@ -32,6 +32,8 @@ import com.google.sps.utils.QuizTimingPropertiesUtils;
 
 public final class UserUtils {
     public static final String SESSION_ID_COOKIE_NAME = "SessionID";
+    public static final int ADDED_POINTS = 20;
+    public static final long ADDED_POINTS_LONG = 20L;
     private static final Logger log = Logger.getLogger(UserUtils.class.getName());
 
     /**
@@ -159,7 +161,7 @@ public final class UserUtils {
 
     QuizTimingPropertiesUtils utils = new QuizTimingPropertiesUtils();
     if(userEntity.getProperty("lastAwardedUploadPoints") == null || utils.isTimestampOutdated((long) userEntity.getProperty("lastAwardedUploadPoints"))){
-        addPoints(userEntity, 20, datastore);
+        addPoints(userEntity, ADDED_POINTS, datastore);
         userEntity.setProperty("lastAwardedUploadPoints", System.currentTimeMillis());
         datastore.put(userEntity);
     }
--- a/src/main/java/com/google/sps/utils/UserUtils.java
+++ b/src/main/java/com/google/sps/utils/UserUtils.java
@@ -32,7 +32,7 @@ import com.google.sps.utils.QuizTimingPropertiesUtils;
 
 public final class UserUtils {
     public static final String SESSION_ID_COOKIE_NAME = "SessionID";
-    public static final long ADDED_POINTS = 20L;
+    public static final int ADDED_POINTS = 20;
     private static final Logger log = Logger.getLogger(UserUtils.class.getName());
 
     /**
@@ -144,9 +144,9 @@ public final class UserUtils {
   /**
     adds a specified number of points to the user's points
   */
-  public static void addPoints(Entity userEntity, long numPoints, DatastoreService datastore){
+  public static void addPoints(Entity userEntity, int numPoints, DatastoreService datastore){
     try {
-      userEntity.setProperty("score", ((Number) userEntity.getProperty("score")).longValue() + numPoints);
+      userEntity.setProperty("score", ((Number) userEntity.getProperty("score")).intValue() + numPoints);
     } catch (NullPointerException e) {
       userEntity.setProperty("score", numPoints);                
     }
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -150,4 +151,18 @@ public static void addPoints(Entity userEntity, int numPoints, DatastoreService
...
     datastore.put(userEntity);
   }
+
+  /**
+    Adds 20 points to a user for uploading if it has been more than a day since they last uploaded
+  */
+  public static void addUploadPoints(Entity userEntity, DatastoreService datastore){
+
+    QuizTimingPropertiesUtils utils = new QuizTimingPropertiesUtils();
+    if(userEntity.getProperty("lastUploadTime") == null || utils.isTimestampOutdated((long) userEntity.getProperty("lastUploadTime"))){
```
colemanliyah, line 161: should they be getting points if they never uploaded?

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -150,4 +151,18 @@ public static void addPoints(Entity userEntity, int numPoints, DatastoreService
...
+        addPoints(userEntity, 20, datastore);
+        userEntity.setProperty("lastUploadTime", System.currentTimeMillis());
```
colemanliyah, line 163: I think the time gets updated when it's checked. I could be mistaken on that but if not then you wouldn't need line 163

src/test/java/com/google/sps/UserUtilTest.java
```diff
@@ -419,4 +419,56 @@ public void addPoints_firstPoints(){
...
+    Entity user1 = new Entity("user");
+    user1.setProperty("userId", "newUser1");
+    datastore.put(user1);
+ 
+    UserUtils.addUploadPoints(user1, datastore);
+ 
+    Entity userEntity = UserUtils.getEntityFromDatastore("user", "userId", "newUser1", datastore);
+    long expected = 20;
+    
+    Assert.assertEquals(expected, userEntity.getProperty("score"));
```
colemanliyah, line 436: You could probably combine these lines and say Assert.assertEquals(20, blah);

src/test/java/com/google/sps/UserUtilTest.java
```diff
@@ -419,4 +419,56 @@ public void addPoints_firstPoints(){
...
+    user1.setProperty("score", 0);
+    user1.setProperty("lastUploadTime", System.currentTimeMillis()); 
+    datastore.put(user1);
+ 
+    UserUtils.addUploadPoints(user1, datastore);
+ 
+    Entity userEntity = UserUtils.getEntityFromDatastore("user", "userId", "newUser1", datastore);
+    long expected = 0;
+    
+    Assert.assertEquals(expected, userEntity.getProperty("score"));
```
colemanliyah, line 454: Same as above and for the one below

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: This function is only called after they upload (if lastUploadTime is null is means it's never gone into that loop before, so that means this is the first time they're uploading)

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: this is the only place this property is being set/updated; I'll rename this variable to be lastAwardedUploadPoints

src/test/java/com/google/sps/UserUtilTest.java
EmmaWenzel: I get an error if I do that, it needs to be casted to long

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -150,4 +151,18 @@ public static void addPoints(Entity userEntity, int numPoints, DatastoreService
...
   }
+
+  /**
+    Adds 20 points to a user for uploading if it has been more than a day since they last uploaded
+  */
+  public static void addUploadPoints(Entity userEntity, DatastoreService datastore){
+
+    QuizTimingPropertiesUtils utils = new QuizTimingPropertiesUtils();
+    if(userEntity.getProperty("lastAwardedUploadPoints") == null || utils.isTimestampOutdated((long) userEntity.getProperty("lastAwardedUploadPoints"))){
+        addPoints(userEntity, 20, datastore);
```
devonwasson, line 162: the "20" should be a named constant. That way, if we ever want to change it, all of our tests won't break. If we use a named constant, we can change all hard coded references to 20 with a reference to the constant, and then we only ever need to change the number in one spot if we want to change it in the future.

src/test/java/com/google/sps/UserUtilTest.java
EmmaWenzel: Alright, that's done -- with the add points function as well

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: Alright I made a long variable called upload points

PR #53, opened 2020-07-23
Add a createUserEntity function to userUtils

hannahsalameh:
(Code refactoring) Made a function that takes in all necessary parameters and creates a User Entity. If a user already exists, it just updates the Session ID (code used to be in PopulateUserUtils.java)

### DIFF
--- a/src/main/java/com/google/sps/utils/UserUtils.java
+++ b/src/main/java/com/google/sps/utils/UserUtils.java
@@ -11,9 +11,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
+ 
 package com.google.sps.utils;
-
+ 
 import java.io.*;
 import javax.servlet.*;
 import javax.servlet.http.*;
@@ -28,11 +28,11 @@ import com.google.appengine.api.datastore.Query.Filter;
 import com.google.appengine.api.datastore.Query.FilterOperator;
 import com.google.appengine.api.datastore.Query.FilterPredicate;
 import java.util.logging.Logger;
-
+ 
 public final class UserUtils {
     public static final String SESSION_ID_COOKIE_NAME = "SessionID";
     private static final Logger log = Logger.getLogger(UserUtils.class.getName());
-
+ 
     /**
     * Returns a single Entity object that can then be used in a servlet. 
     * The entityPropertyTitle and entityPropertyValue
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -11,9 +11,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
+
```
EmmaWenzel, line 14: this happened to both me and Liyah, not sure what's up with it, but you can get rid of all these spaces (like the character space " " not the line) on lines 14, 16, 31, 35, 73, 77, 82, 86, 90, 115, 120, 133, 139, 144, 163

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -47,25 +47,43 @@
     * @param  entityPropertyValue  the value of the property that is filtered for
     * @return                      a single entity that has the same value for the property in the parameter 
     */
```
ofk-goog, line 49: You've added a new function right between the documentation for `getEntityFromDatastore` and the definition of `getEntityFromDatastore`. Please move the new code somewhere else, and give it its own documentation.

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -47,25 +47,43 @@
...
+
+public static Entity initializeUser(String userId, String name, String sessionID, DatastoreService datastore){
+        Entity userEntity = getEntityFromDatastore("user", "userID", userId, datastore);
```
ofk-goog, line 52: You spelled the key "userID" here and "userId" earlier. That's a pretty devious bug, because it'll either cause user data to be overwritten by accident, or prevent user data from being initialized, depending on which spelling is correct.

This is why I've been recommending that you write a `User` class that lets you not have to deal with `Entity` directly. This is my big "I told you so" moment, and I'm not wasting it. See? I'm gloating.

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -47,25 +47,43 @@
...
+        if(userEntity == null){
```
ofk-goog, line 53: Branching like this makes me nervous. The behavior of `initializeUser` is completely different, based on whether the user exists yet or not. That reduces the risk of overwriting user data... but it also means that after calling initializeUser, the caller has no idea what the state of the user is.

I'd rather the caller be responsible for checking whether the user has been initialized yet (ie whether their ID is present in datastore). What do you think?

src/main/java/com/google/sps/utils/UserUtils.java
ofk-goog: Lmao, I misread and there's no bug. Oh well. My point stands. Mostly

src/main/java/com/google/sps/utils/UserUtils.java
hannahsalameh: I think thats fine, plus it makes more sense with what the function is named. I'll change the implementation

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -33,7 +33,35 @@
...
     private static final Logger log = Logger.getLogger(UserUtils.class.getName());
  
-    /**
+   
+     /**
+    * Takes user's information and creates an entity from it. 
+    * This function inputs all the parameters, while also initializing
+    * some parameters to null values. The entity returned can be stored in the
+    * datastore.
+    * If an entity with the same ID exists, it just updates the SessionID of the first entity.
```
ofk-goog, line 42: This is no longer true

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -47,25 +75,24 @@
...
     * @return                      a single entity that has the same value for the property in the parameter 
     */
+    
   public static Entity getEntityFromDatastore(String entityName, String entityPropertyTitle, String entityPropertyValue, DatastoreService datastore) {
 
     if(entityPropertyValue == "" || entityName == "" || entityPropertyTitle == "" || datastore == null){
         return null;
     }
-
+
```
ofk-goog, line 84: Your editor probably has a "remove trailing whitespace" button somewhere. Please find and depress that button.

PR #54, opened 2020-07-24
Filter and Web.xml For Authentication

hannahsalameh:
This is just the filter and web.xml. When it gets merged, it will break the website because logging in does not work yet

### DIFF
--- a/src/main/java/com/google/sps/AuthenticationFilter.java
+++ b/src/main/java/com/google/sps/AuthenticationFilter.java
@@ -1,0 +1,57 @@
+package com.google.sps.filter;
+ 
+import java.io.IOException;
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.sps.utils.UserUtils;
+import com.google.sps.utils.CookieUtils;
+ 
+public class AuthenticationFilter implements Filter {
+    private ServletContext context;
+    private DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+ 
+    public void init(FilterConfig fConfig) throws ServletException {
+        this.context = fConfig.getServletContext();
+        this.context.log("AuthenticationFilter initialized");
+    }
+ 
+    /**
+    * Checks that a session ID stored as a cookie is also attached to a user entity in the database.
+    * If it is not attached, it redirects to the login page.
+    * 
+    * This function should be called every time a user tries to go to a new page. 
+    *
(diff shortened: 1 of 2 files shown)


hannahsalameh:
can still be reviewed 

NOTE TO SELF: dont merge until login merged

src/main/webapp/web.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee
```
devonwasson, line 4: I don't think a lot of the HTML files in this web.xml are named this way yet. We should't be adding this in this PR. Instead, we should just remove references to files that dont exist or are incorrectly named, and then add that code back in during the relevant PRs.

src/main/java/com/google/sps/AuthenticationFilter.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
```
devonwasson, line 16: It's typically not great to import *, just import whatever you need.

src/main/java/com/google/sps/AuthenticationFilter.java
```diff
@@ -0,0 +1,57 @@
...
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.sps.utils.UserUtils;
+import com.google.sps.utils.CookieUtils;
```
devonwasson, line 21: cookieutils is not used and can be safely removed.

src/main/java/com/google/sps/AuthenticationFilter.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
```
devonwasson, line 13: javax.servlet.http.HttpSession is not used, i believe

src/main/java/com/google/sps/AuthenticationFilter.java
```diff
@@ -0,0 +1,57 @@
...
+    */
+ 
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        HttpServletRequest req = (HttpServletRequest) request;
+        HttpServletResponse res = (HttpServletResponse) response;
+ 
+        Cookie cookies[] = req.getCookies();
+        Entity userEntity = UserUtils.getUserFromCookie(cookies, datastore);
+ 
+        //check to make sure there is an active session and that its attached to a person
```
devonwasson, line 47: "its" should be "it's"

src/main/webapp/web.xml
hannahsalameh: discussed offline!

PR #55, opened 2020-07-24
quiz page html and js files

colemanliyah:
The quiz html page is updated with all the needed tags. The js file has all the fetch functions need to populate the quiz page and select the right users to give points to.

### DIFF
--- a/src/main/webapp/quiz.html
+++ b/src/main/webapp/quiz.html
@@ -1,49 +1,38 @@
 <!--gi Music: https://www.bensound.com -->
+ 
 <!DOCTYPE html>
 <html>
   <head>
     <meta charset="UTF-8">
     <title>Quiz Page</title>
-    <style src="style.css"></style>
+    <link rel="icon" href="/images/plantFav.png" type="image/x-icon">
+    <link rel="stylesheet" href="style.css">
     <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css" integrity="sha384-9aIt2nRpC12Uk9gS9baDl411NQApFmC26EwAOH8WgZl5MYYxFfc+NcPb1dKGj7Sk" crossorigin="anonymous">
-    <style>
-        #container {
-            margin-left: auto;
-            margin-right: auto;
-            width: 1500px;
-        }
-    </style>
     <audio controls id="background_music" style="display:none">
         <source src="/images/bensound-littleidea.mp3">
     </audio> 
   </head>
-  <body onload="checkSignIn()" onload="document.getElementById('background_music').play()">
-    <div id="container">
-        </br>
-        </br>
-        <h3 style="margin-left: 650px">Quiz Time!</h3>
-        </br>
-
-        <button type="button" class="btn btn-success btn-lg btn-block" id="start_quiz"> Start Quiz! </button>
-        <div id="quiz_time" style="display: none">
-            <div class="card">
-                <div class="card-body"> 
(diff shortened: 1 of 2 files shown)


src/main/webapp/quizFunctions.js
```diff
@@ -13,51 +13,31 @@ if(document.getElementById("sound").onclick = function() {
...
-
+ 
 document.getElementById("start_quiz").onclick = function() {
-    document.getElementById("quiz_time").style.display = "block";
-    document.getElementById("start_quiz").style.display = "none";
-    generateQuizQuestions();
-}
+    fetch("/user-quiz-status-servlet").then(response => response.json()).then((tasks) => {
+        if(tasks == true){
+            console.log("You already took this quiz!");
```
hannahsalameh, line 20: can u display it on the page instead of console.logging this?

src/main/webapp/quizFunctions.js
```diff
@@ -13,51 +13,31 @@ if(document.getElementById("sound").onclick = function() {
...
+        } else {
+            fetch("/get-user-images").then(response => response.json()).then((players_ids_and_photos) => {
+                Object.keys(players_ids_and_photos).forEach(function(key) {
+                    fetch('/get-image?blobKey=' + players_ids_and_photos[key]).then((pic) => {
+                    let button_for_picture = document.createElement("BUTTON");
```
hannahsalameh, line 25: can u make button lowercase? just so it goes w the rest of the js

src/main/webapp/quiz.html
```diff
@@ -1,49 +1,38 @@
...
+            <h3 id="headings">Quiz Time!</h3>
+            </br>
+ 
+            <button type="button" class="button" id="start_quiz"> Start Quiz! </button>
+            <div id="quiz_time" style="display: none">
+                <div id="questions">
+                </div>      
+                </br>
+                
+                <form action="/answer-quiz-question" method="POST" id="quiz_photos">
```
hannahsalameh, line 30: what is this for? if u want to do it with the javascript, u dont need a form

hannahsalameh: or does this get populated somewhere

colemanliyah: This gets populated to another servlet that will be in another pr. I didn't want to do the form in the js anymore

src/main/webapp/quizFunctions.js
colemanliyah: are you talking about for line 25? I believe that when creating a button element with JS it should be in all caps

src/main/webapp/quizFunctions.js
colemanliyah: Yes, I added an alert here!

src/main/webapp/quizFunctions.js
```diff
@@ -13,51 +13,31 @@ if(document.getElementById("sound").onclick = function() {
...
-}
-
+ 
 document.getElementById("start_quiz").onclick = function() {
-    document.getElementById("quiz_time").style.display = "block";
-    document.getElementById("start_quiz").style.display = "none";
-    generateQuizQuestions();
-}
+    fetch("/user-quiz-status-servlet").then(response => response.json()).then((tasks) => {
+        if(tasks == true){
```
devonwasson, line 19: the variable "tasks" seems to be a bit oddly named. Is it a boolean? We're comparing it to the boolean "true" so I'm assuming it is. If that's the case, we should name it something more descriptive.

colemanliyah: done!

PR #56, opened 2020-07-24
servelt for adding points related to the quiz

colemanliyah:
This servlet calls on helper functions to give points to who completed a quiz and give points to the person voted for in a quiz.


src/main/java/com/google/sps/servlets/QuizPoints.java
```diff
@@ -0,0 +1,66 @@
...
+
+        //Adds points to the user who took the quiz
+        Long quiz_time = QuizTimingPropertiesUtils.getQuizTimestampProperty("Game", "gameID", userEntity.getProperty("currentGame").toString(), datastore);
+        Long user_time = QuizTimingPropertiesUtils.getQuizTimestampProperty("user", "userID", userEntity.getProperty("userID").toString(), datastore);
+        QuizTimingPropertiesUtils.giveUserQuizTakenPoints(QuizTimingPropertiesUtils.userTookQuiz(user_time, quiz_time), userEntity, datastore);        
+
+        //Adds points to the user who got voted for in the quiz 
+        String clicked_user_id = getParameter(request, "user_picture", "");
+        Entity user_clicked = UserUtils.getEntityFromDatastore("user", "userID", clicked_user_id, datastore);
+        user_clicked.setProperty("score", ((Number) user_clicked.getProperty("score")).intValue() + 20);
```
EmmaWenzel, line 53: you can use the add points function for this I think but not necessary so up to you

src/main/java/com/google/sps/servlets/QuizPoints.java
```diff
@@ -0,0 +1,66 @@
...
+        datastore.put(user_clicked);
+
+       response.sendRedirect("/gameBoard.html");
+    }
+
+    private String getParameter(HttpServletRequest request, String name, String defaultValue) {
```
EmmaWenzel, line 59: you could put this in user utils, I think it might be helpful elsewhere

devonwasson: This would actually be really useful in something like a "HttpRequestUtils" file. And then we can add some test code for it as well. This is a nice wrapper and we should use it in more places

src/main/java/com/google/sps/servlets/QuizPoints.java
devonwasson: +1, we should reuse code as much as possible and this is a good way to do it.

devonwasson, line 53: "20" should be a named constant. that way, if we want to change that value, we only have to change it here and not in multiple spots (in case it ends up appearing in multiple spots)

src/main/java/com/google/sps/servlets/QuizPoints.java
```diff
@@ -0,0 +1,66 @@
...
+        //Adds points to the user who took the quiz
+        Long quiz_time = QuizTimingPropertiesUtils.getQuizTimestampProperty("Game", "gameID", userEntity.getProperty("currentGame").toString(), datastore);
+        Long user_time = QuizTimingPropertiesUtils.getQuizTimestampProperty("user", "userID", userEntity.getProperty("userID").toString(), datastore);
+        QuizTimingPropertiesUtils.giveUserQuizTakenPoints(QuizTimingPropertiesUtils.userTookQuiz(user_time, quiz_time), userEntity, datastore);        
+
+        //Adds points to the user who got voted for in the quiz 
+        String clicked_user_id = getParameter(request, "user_picture", "");
+        Entity user_clicked = UserUtils.getEntityFromDatastore("user", "userID", clicked_user_id, datastore);
+        user_clicked.setProperty("score", ((Number) user_clicked.getProperty("score")).intValue() + 20);
+        datastore.put(user_clicked);
```
devonwasson, line 54: shouldn't the datastore already know about this user? Do we need to do a whole new put? we already set their score property in the line above so I'm not sure we need this.

src/main/java/com/google/sps/servlets/QuizPoints.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
+
+//This servlet's request comes from the quiz.html page; the request is the ID of the user who was voted for in a quiz
+//This servelt's functionality adds points the user who took a quiz and adds points the person who is voted for in a quiz
+@WebServlet("/quiz-points")
+public class QuizPoints extends HttpServlet {
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
devonwasson, line 29: I think we want to just declare it here but actually assign it in the constructor.

src/main/java/com/google/sps/servlets/QuizPoints.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.servlet.http.HttpServletResponse;
+import com.google.appengine.api.datastore.FetchOptions;
+import java.text.DateFormat;
+import java.util.Date;
+import com.google.appengine.api.datastore.Key;
+import com.google.sps.QuizTimingPropertiesUtils;
+import com.google.sps.utils.UserUtils;
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
```
devonwasson, line 22: Please double check these imports, I don't think they're all being used (Date and DateFormat, for example). 

Also, any imports where you import * should specifically call out only what you need rather than *.

src/main/java/com/google/sps/servlets/QuizPoints.java
colemanliyah: You set the property with line 53 and then in order for it to be updated you have to put it back into the datastore

src/main/java/com/google/sps/servlets/QuizPoints.java
colemanliyah: I added the helper class for this

src/main/java/com/google/sps/servlets/QuizPoints.java
colemanliyah: I included the user utils add Points function for this

src/main/java/com/google/sps/servlets/QuizPoints.java
colemanliyah: Okay I removed some of the them, I kept the last three with stars because those are used for the cookies

pom.xml
```diff
@@ -2,69 +2,76 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
-
+
```
devonwasson, line 5: Please remove the space characters added to blank lines.

PR #57, opened 2020-07-24
Imagesforquizpage

colemanliyah:
This servlet loads images for the quiz page if the user is logged in. It does not add the photo of the user who is logged in though

### DIFF
--- a/src/main/java/com/google/sps/servlets/getUserImagesForQuizPage.java
+++ b/src/main/java/com/google/sps/servlets/getUserImagesForQuizPage.java
@@ -1,0 +1,55 @@
+package com.google.sps.servlets;
+ 
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Key;
+import com.google.sps.utils.QuizTimingPropertiesUtils;
+import com.google.sps.utils.UserUtils;
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.HashMap;
+ 
+@WebServlet("/get-user-images")
+public class getUserImagesForQuizPage extends HttpServlet {
+ 
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+ 
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        Cookie cookies[] = request.getCookies();
+        if(cookies == null){
+            response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            return;
+        }
+        Entity userEntity = UserUtils.getUserFromCookie(cookies, datastore);
+        if(userEntity == null){
+            response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            return;
+        } 
+ 
+        HashMap<String,String> user_ids_and_pictures = new HashMap<String, String>();
+ 
+        current_game = UserUtils.getEntityFromDatastore("Game", "gameId", (userEntity.getProperty("gameId")).toString(), datastore);
+        for(String playerID : (ArrayList<String>) current_game.getProperty("userIDs")) {
+            if(player.getProperty("userID").equals(playerID) && !user_entity.getProperty("userID").equals(playerID)) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/sps/servlets/getUserImagesForQuizPage.java
```diff
@@ -0,0 +1,56 @@
...
+        if(userEntity == null){
+            response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            return;
+        } 
+ 
+        HashMap<String,String> user_ids_and_pictures = new HashMap<String, String>();
+ 
+        current_game = UserUtils.getEntityFromDatastore("Game", "gameId", (userEntity.getProperty("gameId")).toString(), datastore);
+        for(String playerID : (ArrayList<String>) current_game.getProperty("userIDs")) {
+            if(player.getProperty("userID").equals(playerID) && !user_entity.getProperty("userID").equals(playerID)) {
```
EmmaWenzel, line 47: I don't think player or user_entity have been declared, are these supposed to be userEntity and playerID?

EmmaWenzel: ^ correct me if I'm wrong I might've missed them

src/main/java/com/google/sps/servlets/getUserImagesForQuizPage.java
```diff
@@ -0,0 +1,56 @@
...
+                user_ids_and_pictures.put((player.getProperty("userIds")).toString(), (player.getProperty("blobkey")).toString());
```
EmmaWenzel, line 48: User entities have the property userID (userIds is a property of the game entity)

src/main/java/com/google/sps/servlets/getUserImagesForQuizPage.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.gson.Gson;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Key;
```
EmmaWenzel, line 15: I don't think you're using a handful of these (key, fetch options, query, prepared query, quiz timing properties utils)

src/main/java/com/google/sps/servlets/getUserImagesForQuizPage.java
colemanliyah: the userEntity is declared on line 37, it grabs the current user from the cookies. Player Id is declared in the for loop on line 46

src/main/java/com/google/sps/servlets/getUserImagesForQuizPage.java
colemanliyah: fixed this!

src/main/java/com/google/sps/servlets/getUserImagesForQuizPage.java
colemanliyah: Got rid of some of these

src/main/java/com/google/sps/servlets/getUserImagesForQuizPage.java
EmmaWenzel: one more sorry, I think it's declared as userEntity originally and used as user_entity

colemanliyah: got it!

src/main/java/com/google/sps/servlets/getUserImagesForQuizPage.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.servlet.http.HttpServletResponse;
+import com.google.sps.utils.QuizTimingPropertiesUtils;
+import com.google.sps.utils.UserUtils;
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+ 
+//This servlet loads user images to the quiz page for a logged-in user
```
ofk-goog, line 20: Which images?
Please add something like (my interpretation): "This returns a JSON map from user ID to image blob key for each player in the current game, excluding the logged-in player."

Also, why exclude the logged-in player?

colemanliyah: we don't want the user to be able to vote for their own picture in the quiz, I added documentation to help explain this...let me know if it helps

ofk-goog: What's the chain of logic between "we don't want the logged-in user to vote for their own game" and "therefore, we don't return the user's own picture here"?

I'm worried you're selecting a pair of pictures to judge in the JavaScript code. That would be a security bug

colemanliyah: The idea is that the pictures of each user playing that game will be loaded onto the quiz page except for the one that belongs to the current user

ofk-goog: Oh okay. I thought it was a matter of selecting a pair, but now I understand you display every picture all at once.

PR #58, opened 2020-07-27
Leaderboard work

EmmaWenzel:
Create a User class, create a function that gets all the users in a game and sorts them by score, and then puts them in user objects (setting the score, userId, and userName properties for the leaderboard).


src/main/java/com/google/sps/utils/User.java
```diff
@@ -0,0 +1,57 @@
...
+        return quiz_timing;
+    }
+    public String getBlobKey(){
+        return blobKey;
+    }
+    public String getGameId(){
+        return gameId;
+    }   
+
+    public void setId(String userID){
```
hannahsalameh, line 39: it should not be possible to override the userID --> im not sure exactly what errors u could cause doing this but it would b wild

src/main/java/com/google/sps/utils/User.java
```diff
@@ -0,0 +1,57 @@
...
+        userID = userID;
+    }
+    public void setName(String userName){
```
hannahsalameh, line 42: this too -- u really shouldn't be able to change your name once the user entity is created

src/main/java/com/google/sps/utils/User.java
EmmaWenzel: got it, I took these sets out, they're just part of the constructor now

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -28,7 +28,12 @@
 import com.google.appengine.api.datastore.Query.FilterOperator;
 import com.google.appengine.api.datastore.Query.FilterPredicate;
 import java.util.logging.Logger;
+<<<<<<< HEAD
+import com.google.sps.utils.User;
+import com.google.appengine.api.datastore.Query.SortDirection;
+=======
 import com.google.sps.utils.QuizTimingPropertiesUtils;
+>>>>>>> a4f50e37510867b1aba9cb0c5a74f93d45725139
```
ofk-goog, line 36: This is a merge conflict you need to clean up

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -153,6 +158,29 @@ public static void addPoints(Entity userEntity, int numPoints, DatastoreService
...
+    
+    // stores each user in a User object 
+    ArrayList<User> users = new ArrayList<>();
+    for (Entity entity : results.asIterable()) {
+        String userID = (String) entity.getProperty("userID");
+        String userName = (String) entity.getProperty("username");
+        int score = ((Number) entity.getProperty("score")).intValue();
+    
+        User user = new User(userID, userName, score);
+        users.add(user);
```
ofk-goog, line 178: Most of the value of having a `User` class is to encapsulate the user `Entity` object. If you have a `User` but there's code outside of `User` to convert `User` to and from `Entity`, then `User` is not making your life any easier.

I recommend an approach where an `Entity` is the sole private variable inside `User`. Then you can create a `User` around an entity and manipulate the entity by calling the setters and getters on `User`.

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: got it, sorry about that

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: Alright cool I changed most of it to fit the syntax of your code for it; with the exception of a few extra parameters to be gotten by the javascript for the leaderboard

src/main/java/com/google/sps/utils/User.java
```diff
@@ -0,0 +1,102 @@
...
+      throw new IllegalArgumentException(
+          "Attempted to interpret Entity of kind " + userEntity.getKind()
+          + " as a User.");
+    }
+    entity = userEntity;
+
+    // set values for the leaderboard
+    userID = (String) entity.getProperty("userID");
+    userName = (String) entity.getProperty("userName");
+    score = ((Number) entity.getProperty("score")).intValue();
```
ofk-goog, line 39: These are already stored in the Entity. Do you need to store them outside as well? Is it intentional that the getters return these value, but the setters don't change these values? If so, they should probably have different names to avoid confusion.

src/main/java/com/google/sps/utils/User.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  // current game score
+  public int getScore(){
+    return score;
+  }
+  public void setScore(int score){
+    entity.setProperty("score", score);
+    datastore.put(entity);
+  }
```
ofk-goog, line 82: Unless the current behavior is intentional, I recommend adding a unit test for each getter and setter pair, to make sure the getter gets what the setter sets.
```
User user = new User(new Entity(User.USER_ENTITY_KIND));
int expected = 31;
user.setScore(expected);
Assert.assertEquals(expected, user.getScore());
```

src/main/java/com/google/sps/utils/User.java
```diff
@@ -0,0 +1,102 @@
...
+  
+/**
+ * Wrapper around a datastore Entity representing a user.
+ *
+ * This adds some type safety and relieves the need for type casting when
+ * handling user entities.
+ */
+public final class User {
+
+  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
ofk-goog, line 16: It'd be better not to have this class automatically update datastore with each setter call. A few reasons:

* It'll be easier to unit test if it doesn't use datastore.
* Often you'll want datastore updates to be all-or-nothing. For example, imagine you want to add the user to a new game and set their score to zero. If you use two separate calls to datastore, something could go wrong in between, and then they'd be in the new game with a nonzero score.
* If you change several fields at once, it'll be much slower if you're sending a separate request to datastore for each field.

Instead, I recommend using the pattern `datastore.put(user.getEntity())` in the code that uses `User`.

EmmaWenzel: That makes sense, I took datastore out.

src/main/java/com/google/sps/utils/User.java
EmmaWenzel: Got it; I made tests for each setter and getter.

src/main/java/com/google/sps/utils/User.java
EmmaWenzel: It was intentional to have them stored outside of the entity, because these properties are displayed on the leader board, so they need to be accessible from the javascript. I did change the setters to also set these values, though.

ofk-goog: We talked "in person" and decided to get rid of the extra local variables. We don't need them unless we use GSON to convert `User` to JSON. And it's simpler to not use GSON, because GSON is not well matched to our use case.

PR #59, opened 2020-07-27
Leaderboard html js

EmmaWenzel:
Add js functions to get users of a game from the server and create a list element to list them by score for the leaderboard.

### DIFF
--- a/src/main/webapp/gameBoard.html
+++ b/src/main/webapp/gameBoard.html
@@ -41,9 +41,7 @@
             </br>
  
             <div id="leaderboard-container" onload = "loadUsers()">
-                <body onload="loadUsers()">
-                    <ol id="leader-board"></ol>
-                </body>
+                <ol id="leader-board"></ol>
             </div>
             </br>
  
--- a/src/main/webapp/gameBoard.html
+++ b/src/main/webapp/gameBoard.html
@@ -40,8 +40,10 @@
             <h5 id="headings"> Game Name </h5>
             </br>
  
-            <div id="leaderboard-container">
-                LeaderBoard Elements will be populated Here
+            <div id="leaderboard-container" onload = "loadUsers()">
+                <body onload="loadUsers()">
+                    <ol id="leader-board"></ol>
+                </body>
             </div>
             </br>
  
--- a/src/main/webapp/gameFunctions.js
+++ b/src/main/webapp/gameFunctions.js
@@ -5,3 +5,42 @@ function getUserQuizStatus() {
         }
     });
 }
+
+/** 
+* fetches users of the game from the server 
+* and adds score, id and name to the page
+*/
+function loadUsers() {
+    fetch('/populate-leaderboard').then(response => response.json()).then((users) => {
+        const userListElement = document.getElementById('leader-board');
+        console.log(users[0]);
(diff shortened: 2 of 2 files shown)


src/main/webapp/gameBoard.html
```diff
@@ -40,8 +40,10 @@
             <h5 id="headings"> Game Name </h5>
             </br>
  
-            <div id="leaderboard-container">
-                LeaderBoard Elements will be populated Here
+            <div id="leaderboard-container" onload = "loadUsers()">
+                <body onload="loadUsers()">
+                    <ol id="leader-board"></ol>
+                </body>
```
colemanliyah, line 46: Do you need this body tag? This file already has one so maybe you could just put the loadUsers() onload function there

colemanliyah, line 46: It looks like you are calling "loadUsers()" twice in a row, did you mean to do this?

src/main/webapp/gameFunctions.js
```diff
@@ -5,3 +5,39 @@ function getUserQuizStatus() {
...
+        }
+    });
+}
+ 
+/** Creates a list item that contains the user */
+function createUserElement(user){
+  
+  // create list elements
+  const userElement = document.createElement('li');
+  const userIDElement = document.createElement('div');
```
colemanliyah, line 27: Do you need line 27 calling the ID or should we just show the username?

src/main/webapp/gameFunctions.js
```diff
@@ -5,3 +5,39 @@ function getUserQuizStatus() {
...
+
+/** 
+* fetches users of the game from the server 
+* and adds score, id and name to the page
+*/
+function loadUsers() {
+    fetch('/populate-leaderboard').then(response => response.json()).then((users) => {
+        const userListElement = document.getElementById('leader-board');
+        for(i = 0; i < users.length; i++){
+            userListElement.appendChild(createUserElement(users[i]));
```
colemanliyah, line 17: since you already have to loop through all the users, can we compare their scores too so that we can put them on the leaderboard in order?

src/main/webapp/gameFunctions.js
EmmaWenzel: I figured we'd have both because 2 users can have the same name

src/main/webapp/gameFunctions.js
EmmaWenzel: It's put in order of score by the servlet

src/main/webapp/gameBoard.html
EmmaWenzel: I removed it

EmmaWenzel: ^ removed the body tag and call

src/main/webapp/gameFunctions.js
```diff
@@ -5,3 +5,39 @@ function getUserQuizStatus() {
         }
     });
 }
+
+/** 
+* fetches users of the game from the server 
+* and adds score, id and name to the page
+*/
+function loadUsers() {
```
ofk-goog, line 13: Would it be more accurate to call this something like `loadLeaderboard`?

PR #61, opened 2020-07-27
created createSessionIDCookie() fxn (+tests!)

hannahsalameh:
Created a function that takes in the current session and the new session Id and returns a cookie.

### DIFF
--- a/src/main/java/com/google/sps/utils/CookieUtils.java
+++ b/src/main/java/com/google/sps/utils/CookieUtils.java
@@ -53,4 +53,13 @@ public final class CookieUtils {
         }
         return null;
     }
+    public static Cookie createSessionIDCookie(HttpSession oldSession, String newSessionID){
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/utils/CookieUtils.java
```diff
@@ -53,4 +53,13 @@ public static Cookie getCookieFromName(Cookie cookies[], String name){
         }
         return null;
     }
+    public static Cookie createSessionIDCookie(HttpSession oldSession, String newSessionID){
+        if(oldSession != null){
+            oldSession.invalidate();
+        }
+        Cookie cookie = new Cookie("SessionID", newSessionID);
```
ofk-goog, line 60: There's a constant for "SessionID" somewhere, right? Maybe in UserUtils. It'd be good to use that constant here.

PR #62, opened 2020-07-27
created the entire login servlet

hannahsalameh:
This servlet:
  1. checks the validity of the google API token
  2. creates a session ID cookie and sends it to the front end
  3. creates a user entity and sends it to the datastore
  4. redirects to the home page


src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,95 @@
...
+public class LoginServlet extends HttpServlet {
+DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+     
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    JsonFactory jsonFactory = new JacksonFactory();
+    HttpTransport transport = new NetHttpTransport(); 
+
+    GoogleIdTokenVerifier verifier = new GoogleIdTokenVerifier.Builder(transport, jsonFactory)
+            .setAudience(Collections.singletonList("610251294652-9ojdhjhh8kpcdvdbmrvp9nkevgr2n2d8.apps.googleusercontent.com"))
```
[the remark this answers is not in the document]
hannahsalameh: the token can only be used on two websites -- the localhost:8080 and the plantasy launched site. I'm not sure it needs to be hidden since it doesn't work on any other sites.

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,95 @@
...
+            .build();
+
+    Enumeration<String> headers = request.getHeaders("idtoken");
+    String idTokenString = null;
+    while(headers.hasMoreElements()){
+        idTokenString = headers.nextElement();
+    }
+
+    try{
+        GoogleIdToken idToken = verifier.verify(idTokenString); //line where it breaks
```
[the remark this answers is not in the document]
hannahsalameh: whoops! no nothing breaks now

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,95 @@
...
+@WebServlet("/login")
+public class LoginServlet extends HttpServlet {
+DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+     
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    JsonFactory jsonFactory = new JacksonFactory();
+    HttpTransport transport = new NetHttpTransport(); 
+
+    GoogleIdTokenVerifier verifier = new GoogleIdTokenVerifier.Builder(transport, jsonFactory)
```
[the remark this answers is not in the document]
hannahsalameh: you only need to build the verifier once. I can move the verifier call to before the post request

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,95 @@
...
+        } else {
+            userEntity.setProperty("SessionID",sessionID);
+            datastore.put(userEntity);
+        }
+        } else {
+          System.out.println("Invalid ID token.");
+       }
+    } catch(Exception e){
+        e.printStackTrace();
+    }
```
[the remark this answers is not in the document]
hannahsalameh: this catch is just to check that the you can get the GoogleIdToken by verifying the idString. the rest can be moved out of the try/catch I'm pretty sure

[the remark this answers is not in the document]
colemanliyah: +1 to this, I usually have just one line of could in the try block

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -0,0 +1,97 @@
...
+
+          Payload payload = idToken.getPayload();
+          //get user info for user entity
+          String userId = payload.getSubject();
+          String username = (String) payload.get("name");
+
+          //send user data to datastore
+          Entity userEntity = UserUtils.getEntityFromDatastore("user", "userID", userId, datastore);
+          if(userEntity == null){
+            datastore.put(UserUtils.initializeUser(userId, username, sessionID));
```
colemanliyah, line 85: Do we need to check if this function returns null? Or is this already in a test for UserUtils

hannahsalameh: yea! i refactored it a little so now the initialize user fxn just creates the user. you now have to check if the user exists before putting it into the database.

src/main/java/com/google/sps/servlets/LoginServlet.java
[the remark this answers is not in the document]
devonwasson: hmm... having it work on localhost:8080 is risky because anyone could spin up a local web server and use it. Additionally, it is best practice to obscure tokens like this, regardless of where they can be used. You should consider putting it into a json file, then adding that json file to the .gitignore file. That way, you can still use it and build with it but it wont be uploaded to git for the world to see.

src/main/java/com/google/sps/servlets/LoginServlet.java
```diff
@@ -62,26 +62,28 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     String idTokenString = null;
     while(headers.hasMoreElements()){
         idTokenString = headers.nextElement();
+         //header passed in should always have ONLY 1 element (just the id token)
+         //therefore, id token should always be the last element.
```
devonwasson, line 66: We don't know that'll always be true. The website might change in the future and you might want to send more headers to the servlet. Is there any way we can check and be sure that we're getting the id token and not anything else without relying on position?

hannahsalameh: ok (sorry its been a sec since i wrote this code) I just realized that by doing request.getHeaders("idtoken") its always just going to return the value of the header labelled idtoken. U have to still iterate through it, but I'll update the comment to reflect this

src/main/java/com/google/sps/servlets/LoginServlet.java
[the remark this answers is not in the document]
hannahsalameh: just realized something, you can't initialize the GoogleIdToken to be valueless, so everything that relies on it must be inside the try block with it. Unless someone has another suggestion for how I can use it?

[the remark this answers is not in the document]
devonwasson: when you catch, you just return early.

.gitignore
```diff
@@ -1 +1,2 @@
 target/
+/APIKey.txt
```
devonwasson, line 2: I dont think you need the leading / here. It might actually cause problems. if APIKey.txt is in the same level as .gitignore, we can remove the leading /

src/main/java/com/google/plantasy/utils/PasswordReaderUtils.java
```diff
@@ -0,0 +1,44 @@
...
+        File file;
+        try{
+            file = new File(fileLocation);
+        } catch(Exception e){
+            log.severe("Could not find file at location: " + fileLocation);
+            return "noKey";
+        }
+        FileReader fileReader = new FileReader(file);
+        BufferedReader br = new BufferedReader(fileReader); 
+        String apiKey = "noKey";
```
devonwasson, line 37: We can define this right after line 28, and then we can just return apiKey on line 33 rather than hard coding this string twice.

src/test/java/com/google/plantasy/PasswordReaderUtilsTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.mockito.Mock;
+import java.io.IOException;
+import org.mockito.MockitoAnnotations;
+import static org.mockito.Mockito.when;
+import com.google.sps.utils.PasswordReaderUtils;
+ 
+@RunWith(JUnit4.class)
+public final class PasswordReaderUtilsTest{
+ 
+    //Test where the .txt file has more than 1 line of code
```
devonwasson, line 38: Can we add a test where the file doesn't exist?

hannahsalameh: yes i can

src/main/java/com/google/plantasy/utils/PasswordReaderUtils.java
hannahsalameh: I literally have no clue why I did that. Thank u for catching it though!

PR #64, opened 2020-07-28
Renaming Variables

colemanliyah:
changing the name of variables that caused errors when deployed

### DIFF
--- a/src/main/java/com/google/sps/servlets/QuizPoints.java
+++ b/src/main/java/com/google/sps/servlets/QuizPoints.java
@@ -1,5 +1,5 @@
 package com.google.sps.servlets;
- 
+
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/QuizPoints.java
```diff
@@ -1,5 +1,5 @@
 package com.google.sps.servlets;
-
```
devonwasson, line 2: Please remove these extra spaces that get added to files. I have no idea where they keep coming from (because I highly doubt you guys are doing this intentionally) but it might be worth trying to figure out the source of them so we can prevent them from sneaking into future PRs.

colemanliyah: Got it!

PR #65, opened 2020-07-28
game class

colemanliyah:
made a game class and used it in test functions

### DIFF
--- a/src/main/java/com/google/sps/utils/Game.java
+++ b/src/main/java/com/google/sps/utils/Game.java
@@ -5,63 +5,62 @@ import java.util.List;
 import java.util.ArrayList;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
- 
+
 public final class Game {
     DatastoreService datastore;
     private Entity entity;
- 
+
     public Game() {
         datastore = DatastoreServiceFactory.getDatastoreService();
         entity = new Entity("Game");
     }
- 
+
+    //Sets the individual game Id for a game entity
     public void setGameId(String gameId) {
         entity.setProperty("gameId", gameId);
-        datastore.put(entity);
     }
- 
+
+    //Sets the game name for the game entity
     public void setGameName(String game_name) {
         entity.setProperty("gameName", game_name);
-        datastore.put(entity);
     }
- 
+
+    //sets the quiz question for the game entity
     public void setQuizQuestion(String quiz_question) {
         entity.setProperty("quizQuestion", quiz_question);
-        datastore.put(entity);
     }
- 
+
+    //sets the timestamp of the quiz in the game entity 
     public void setQuizTimestamp(long quiz_timestamp) {
         entity.setProperty("quiz_timestamp", quiz_timestamp);
-        datastore.put(entity);
     }
- 
+
+    //sets a list of users that are apart of a game 
     public void setUserIds(ArrayList<String> userIds) {
         entity.setProperty("userIds", userIds);
-        datastore.put(entity);
     }
- 
+
     public Entity getGameEntity() {
         return entity;
     }
- 
+
     public String getGameId() {
         return (String) entity.getProperty("gameId");
     }
- 
+
     public String getGameName() {
         return (String) entity.getProperty("gameName");
     }
- 
+
     public String getQuizQuestion() {
         return (String) entity.getProperty("quizQuestion");
     }
- 
+
     public long getQuizTimestamp() {
         return (long) entity.getProperty("quiz_timestamp");
     }
- 
+
     public ArrayList<String> getUserIds() {
         return (ArrayList<String>) entity.getProperty("userIds");
     }
- 
 }
--- a/src/main/java/com/google/sps/utils/Game.java
+++ b/src/main/java/com/google/sps/utils/Game.java
@@ -1,18 +1,17 @@
 package com.google.sps.utils;
 import com.google.appengine.api.datastore.Entity;
 import java.lang.IllegalArgumentException;
-import java.util.List;
 import java.util.ArrayList;
-import com.google.appengine.api.datastore.DatastoreService;
-import com.google.appengine.api.datastore.DatastoreServiceFactory;
 
 public final class Game {
-    DatastoreService datastore;
+    public final static String game_kind = "Game";
     private Entity entity;
 
-    public Game() {
-        datastore = DatastoreServiceFactory.getDatastoreService();
-        entity = new Entity("Game");
+    public Game(Entity game_entity) {
+        if (game_entity.getKind().equals(game_kind)) {
+            throw new IllegalArgumentException("Attempted to interpret Entity of kind " + game_entity.getKind()+ " as a Game.");
+        }
+    entity = game_entity;
     }
 
     //Sets the individual game Id for a game entity
@@ -63,4 +62,5 @@ public final class Game {
     public ArrayList<String> getUserIds() {
         return (ArrayList<String>) entity.getProperty("userIds");
     }
+
 }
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/utils/Game.java
```diff
@@ -0,0 +1,67 @@
...
+    private Entity entity;
+ 
+    public Game() {
+        datastore = DatastoreServiceFactory.getDatastoreService();
+        entity = new Entity("Game");
+    }
+ 
+    public void setGameId(String gameId) {
+        entity.setProperty("gameId", gameId);
+        datastore.put(entity);
```
EmmaWenzel, line 20: You can take out the datastore.put() calls (and datastore in general) and we'll just add it to datastore where the user object is called, that way it requires less calls to this

src/main/java/com/google/sps/utils/Game.java
```diff
@@ -0,0 +1,67 @@
+package com.google.sps.utils;
+import com.google.appengine.api.datastore.Entity;
+import java.lang.IllegalArgumentException;
+import java.util.List;
+import java.util.ArrayList;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+ 
+public final class Game {
```
EmmaWenzel, line 9: I would maybe make a test for all the setters and getters

src/main/java/com/google/sps/utils/Game.java
```diff
@@ -0,0 +1,67 @@
...
+        entity.setProperty("gameName", game_name);
+        datastore.put(entity);
+    }
+ 
+    public void setQuizQuestion(String quiz_question) {
+        entity.setProperty("quizQuestion", quiz_question);
+        datastore.put(entity);
+    }
+ 
+    public void setQuizTimestamp(long quiz_timestamp) {
```
EmmaWenzel, line 33: you might want to add a short description of each of these properties

src/main/java/com/google/sps/utils/Game.java
colemanliyah: done!

src/main/java/com/google/sps/utils/Game.java
colemanliyah: got it!

src/main/java/com/google/sps/utils/Game.java
colemanliyah: Alright, I made one for this!

src/test/java/com/google/sps/TimingPropertiesTest.java
```diff
@@ -119,32 +120,32 @@ public void userTookQuiz_nullParameters() {
...
- 
+
     @Test
     //Tests valid string, timestamp paramters 
     public void userTookQuiz_validParameters() {
         QuizTimingPropertiesUtils timing_properties_test = new QuizTimingPropertiesUtils();
         
         Entity user = new Entity("user");
-        Entity game = new Entity("game");
+        Game game = new Game();
```
ofk-goog, line 130: yessssssssssssssssssssssssssssssss
(no action required)

src/test/java/com/google/sps/GameTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test 
+    //Tests if the properties are set right for the game class 
+    public void SetsGameEntityCorrectly() {
+        Game game = new Game();
+        game.setGameId("1234");
+        game.setGameName("Plants");
+        game.setQuizQuestion("Which plant looks the healthiest?");
+        game.setQuizTimestamp(1594309443653L);
+        ArrayList<String> users = new ArrayList<String>();
```
ofk-goog, line 53: Please use a non-empty list, so it's clear that `game.getUserIds()` isn't returning a default value.

src/test/java/com/google/sps/GameTest.java
```diff
@@ -0,0 +1,76 @@
...
+    @Test
+    //Tests if the entity created by the game class gets put into datastore
+    public void putInDatastore() {
+        Game game = new Game();
+        game.setGameId("1234");
+
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        datastore.put(game.getGameEntity());
+
+        Assert.assertEquals(UserUtils.getEntityFromDatastore("Game", "gameId", game.getGameId(), datastore), game);
```
ofk-goog, line 74: Here we're comparing an `Entity` to a `Game` with the `.equals()` method of ([I think](https://stackoverflow.com/questions/13607597/does-assertequalsobject-o1-object-o2-uses-the-equals-method)) the `Entity`. I'm not sure why this test passes. (Does it?)

To avoid the whole issue, consider asserting that the game IDs are equal.

src/main/java/com/google/sps/utils/Game.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.List;
+import java.util.ArrayList;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+
+public final class Game {
+    DatastoreService datastore;
+    private Entity entity;
+
+    public Game() {
```
ofk-goog, line 13: Oh, one more thing. I recommend following the same pattern as for the `User` class -- take the `Entity` as a parameter to the constructor. That way you can create a `Game` from an `Entity` retrieved from datastore.

(Checking that `entity.getKind().equals("Game")` is also a good idea)

src/test/java/com/google/sps/TimingPropertiesTest.java
colemanliyah: :)

src/test/java/com/google/sps/GameTest.java
colemanliyah: did this!

src/test/java/com/google/sps/GameTest.java
colemanliyah: Underneath line 53 I add users to this list so that it is not empty is that what you mean?

src/main/java/com/google/sps/utils/Game.java
colemanliyah: Okay I added a constructor for this

src/main/java/com/google/sps/utils/Game.java
```diff
@@ -1,18 +1,17 @@
...
 import com.google.appengine.api.datastore.Entity;
 import java.lang.IllegalArgumentException;
-import java.util.List;
 import java.util.ArrayList;
-import com.google.appengine.api.datastore.DatastoreService;
-import com.google.appengine.api.datastore.DatastoreServiceFactory;
 
 public final class Game {
-    DatastoreService datastore;
+    public final static String game_kind = "Game";
```
ofk-goog, line 7: Please give constants UPPERCASE_NAMES. [Constants are meant to be yelled](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names)

src/test/java/com/google/sps/GameTest.java
```diff
@@ -45,7 +47,8 @@ public void tearDown() {
     @Test 
     //Tests if the properties are set right for the game class 
     public void SetsGameEntityCorrectly() {
-        Game game = new Game();
+        Entity game_entity = new Entity("Game");
+        Game game = new Game(game_entity);
```
ofk-goog, line 51: This can be one line, `Game game = new Game(new Entity(Game.GAME_KIND));`

src/main/java/com/google/sps/utils/Game.java
```diff
@@ -4,11 +4,11 @@
 import java.util.ArrayList;
 
 public final class Game {
-    public final static String game_kind = "Game";
+    public final static String GAME_KIND = "Game";
     private Entity entity;
 
     public Game(Entity game_entity) {
-        if (game_entity.getKind().equals(game_kind)) {
+        if (game_entity.getKind().equals(GAME_KIND)) {
```
ofk-goog, line 11: Ooh you're gonna want to negate this

PR #67, opened 2020-07-29
Send user objects to javascript

EmmaWenzel:
Manually converting (as opposed to using gson) the user objects to json to send to the js so the js doesn't have to deal with entities and entity properties.

### DIFF
--- a/src/main/java/com/google/sps/servlets/populateLeaderboardServlet.java
+++ b/src/main/java/com/google/sps/servlets/populateLeaderboardServlet.java
@@ -29,6 +29,8 @@ import com.google.sps.utils.UserUtils;
 import java.io.*;
 import javax.servlet.*;
 import javax.servlet.http.*;
+import org.json.simple.JSONObject;
+import org.json.simple.JSONArray;
  
 @WebServlet("/populate-leaderboard")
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/populateLeaderboardServlet.java
```diff
@@ -57,11 +59,21 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         return;
     }
  
-    ArrayList<User> users = UserUtils.userList((String) userEntity.getProperty("gameId"), datastore);
+    ArrayList<User> userObjects = UserUtils.userList((String) userEntity.getProperty("gameId"), datastore);
+    
+    // translate to JSON for loadLeaderBoard function
+    JSONArray users = new JSONArray();
+    for(User user : userObjects){
+      JSONObject obj = new JSONObject();
```
colemanliyah, line 67: We talked about this already but I mentioned that you could use a java datastructure here and send it through GSON to the javascript but a JSONArray will work too

ofk-goog: Okay but I love not using GSON. Not using GSON is maybe my favorite part of using JSON.

PR #70, opened 2020-07-29
Minor Changes (I am not creative enough to diversify this pr names!)

colemanliyah:
minor name and syntax changes

### DIFF
--- a/src/main/java/com/google/sps/servlets/QuizPoints.java
+++ b/src/main/java/com/google/sps/servlets/QuizPoints.java
@@ -25,7 +25,7 @@ public class QuizPoints extends HttpServlet {
 
     @Override
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-
+ 
         Cookie cookies[] = request.getCookies();
         if(cookies == null){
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/QuizPoints.java
```diff
@@ -25,7 +25,7 @@
 
     @Override
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-
+
```
devonwasson, line 28: Please remove extra space characters here and on line 49

src/main/webapp/quizFunctions.js
```diff
@@ -23,10 +23,10 @@ document.getElementById("start_quiz").onclick = function() {
                 Object.keys(players_ids_and_photos).forEach(function(key) {
                     fetch('/get-image?blobKey=' + players_ids_and_photos[key]).then((pic) => {
                     let button_for_picture = document.createElement("BUTTON");
-                    button_for_picture.name="user_picture";
+                    button_for_picture.name= "user_picture";
```
devonwasson, line 26: if we're going to add the space on the right (which we should), we should add it on the right also.

PR #72, opened 2020-07-30
Remove null checks

EmmaWenzel:
Remove null checks in game utils functions and some user utils functions and test for those functions throwing exceptions instead.

### DIFF
--- a/src/main/java/com/google/sps/servlets/startGameServlet.java
+++ b/src/main/java/com/google/sps/servlets/startGameServlet.java
@@ -45,6 +45,11 @@ public class startGameServlet extends HttpServlet {
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
  
     String gameName = request.getParameter("game-name");
+
+    if(gameName == ""){
+        response.sendRedirect("/start.html");
+        return;
+    }
     
     // query user 
     Cookie cookies[] = request.getCookies();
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/sps/GameUtilTest.java
```diff
@@ -61,9 +61,12 @@ public void createGame_EmptyGameName() {
...
 
-    Entity actual = GameUtils.createGameEntity("game", null);
-
-    Assert.assertEquals(null, actual);
+    Entity actual = null;
+    try{
+        actual = GameUtils.createGameEntity("game", null);
+    }catch(NullPointerException e){
+        Assert.assertEquals(null, actual);
+    }
```
devonwasson, line 69: here (and everywhere else in this PR) we'll want to actually assert that the exception was thrown.

What you did is technically a correct way of accomplishing this goal, but it's not intuitive and there is a more elegant way to do this:

https://howtodoinjava.com/junit5/expected-exception-example/

Let me know if you have any questions!

src/main/java/com/google/sps/servlets/startGameServlet.java
```diff
@@ -45,6 +45,11 @@ public startGameServlet(){
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
  
     String gameName = request.getParameter("game-name");
+
+    if(gameName == ""){
```
devonwasson, line 49: Might want to change to gameName.isEmpty()

src/test/java/com/google/sps/GameUtilTest.java
EmmaWenzel: Okay, that's done.

src/main/java/com/google/sps/servlets/startGameServlet.java
EmmaWenzel: Got it; changed.

PR #73, opened 2020-07-30
Edit functions for wrapper classes

EmmaWenzel:
Every time get or set property is called, change that to use the user or game class wrappers (plus a few extra changes to fix bugs) and add get/set for the lastAwardedUploadPoints property.


src/main/java/com/google/sps/utils/GameUtils.java
```diff
@@ -42,32 +43,32 @@ public static Entity createGameEntity(String gameName, DatastoreService datastor
...
-    Entity entity = null;
     try{
-      entity = datastore.get(gameKey);
+      gameEntity = datastore.get(gameKey);
 
       String key = KeyFactory.keyToString(gameKey);
-      entity.setProperty("gameId", key);
-      datastore.put(entity);
+      game.setGameId(key);
+      datastore.put(gameEntity);
```
colemanliyah, line 65: You should try to have online one, max two, lines of code in a try block

colemanliyah: Also aren't we putting try catch blocks that test for null in the test classes?

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -163,20 +165,22 @@ public static boolean addBlobKey(String blobKey, Entity userEntity, DatastoreSer
...
   /**
     adds a specified number of points to the user's points
   */
   public static void addPoints(Entity userEntity, int numPoints, DatastoreService datastore){
+    User user = new User(userEntity);
+
     try {
-      userEntity.setProperty("score", ((Number) userEntity.getProperty("score")).intValue() + numPoints);
+      user.setScore(user.getScore() + numPoints);
     } catch (NullPointerException e) {
```
colemanliyah, line 182: Same question as above with the null catches

src/main/java/com/google/sps/utils/UserUtils.java
```diff
@@ -206,10 +210,10 @@ public static void addPoints(Entity userEntity, int numPoints, DatastoreService
   public static void addUploadPoints(Entity userEntity, DatastoreService datastore){
 
     QuizTimingPropertiesUtils utils = new QuizTimingPropertiesUtils();
-    if(userEntity.getProperty("lastAwardedUploadPoints") == null || utils.isTimestampOutdated((long) userEntity.getProperty("lastAwardedUploadPoints"))){
+    User user = new User(userEntity);
+    if(user.getLastAwardedUploadPoints() == 0L || utils.isTimestampOutdated(user.getLastAwardedUploadPoints())){
```
colemanliyah, line 214: So for QuizTimePropertiesUtils you shouldn't have to create an instance of it because all the methods are static.

src/main/java/com/google/sps/utils/GameUtils.java
EmmaWenzel: I have this one here because this is where the potential null would be created (as opposed to be created elsewhere and passed in), so I'm checking for it here

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: Same issue as above ^ if it's the first time the user is getting points, getScore throws a null pointer exception, but we still want to add points so that's what the catch block is for

src/main/java/com/google/sps/utils/UserUtils.java
EmmaWenzel: Okay cool, that's changed

src/main/java/com/google/sps/utils/GameUtils.java
colemanliyah: Okay but can we still reduce it to be only one-two lines in the try block?

EmmaWenzel: Done, sorry!

src/main/java/com/google/plantasy/utils/Game.java
```diff
@@ -8,7 +8,7 @@
     private Entity entity;
 
     public Game(Entity game_entity) {
-        if (game_entity.getKind().equals(GAME_KIND)) {
+        if (!(game_entity.getKind().equals(GAME_KIND))) {
```
devonwasson, line 11: You shouldn't need the extra set of (), just the ! should sufice.

src/main/java/com/google/plantasy/utils/GameUtils.java
```diff
@@ -42,32 +43,31 @@ public static Entity createGameEntity(String gameName, DatastoreService datastor
...
     }catch(EntityNotFoundException e){
       log.severe("EntityNotFoundException; game entity not found when adding game id in create game");
       return null;
     }
+    String key = KeyFactory.keyToString(gameKey);
+    game.setGameId(key);
+    datastore.put(gameEntity);
  
-    return entity;
+    return gameEntity;
```
devonwasson, line 70: considering we have a nice wrapper class around game entity now, we should maybe return the Game rather than the Entity. and if we really want to return an entity, we should be calling game.GetGameEntity() instead

src/main/java/com/google/plantasy/utils/GameUtils.java
```diff
@@ -42,32 +43,31 @@ public static Entity createGameEntity(String gameName, DatastoreService datastor
...
 
     // getting the game id for the game id property
-    Entity entity = null;
     try{
-      entity = datastore.get(gameKey);
-
-      String key = KeyFactory.keyToString(gameKey);
-      entity.setProperty("gameId", key);
-      datastore.put(entity);
+      gameEntity = datastore.get(gameKey);
```
devonwasson, line 61: I'm a bit confused... why do we need to do this? It looks like we only get the entity just to put it back into the datastore on line 68. Do we need to get it at all?

src/main/java/com/google/plantasy/utils/GameUtils.java
```diff
@@ -86,16 +86,17 @@ public static boolean addUserToGame(String userId, Entity gameEntity, DatastoreS
...
         return false;
     }
+
+    Game game = new Game(gameEntity);
  
-    ArrayList<String> userIds = (ArrayList<String>) gameEntity.getProperty("userIds");
+    ArrayList<String> userIds = game.getUserIds();
     if(userIds == null){
       userIds = new ArrayList<String>();
+      game.setUserIds(userIds);
```
devonwasson, line 95: We set the user Ids but we don't update the datastore, so if the server gets refreshed or if we try to reference this again, we'll lose that data. We can call game.GetGameEntity() below and do datastore.put() with that as the param.

src/main/java/com/google/plantasy/utils/UserUtils.java
```diff
@@ -54,13 +54,15 @@ public static Entity initializeUser(String userId, String name, String sessionID
...
         long initialTime = 0L;
+        long initialUploadPoints = 0L;
         Entity userEntity = new Entity("user");
         userEntity.setProperty("username",name);
         userEntity.setProperty("userID",userId);
         userEntity.setProperty("quiz_timing",initialTime);
         userEntity.setProperty("gameId", "");
         userEntity.setProperty("blobKey", null);
         userEntity.setProperty("score", initialScore);
+        userEntity.setProperty("lastAwardedUploadPoints", initialUploadPoints);
```
devonwasson, line 65: we should be able to combine lines 57 and 65. It's good for number to be named if we plan on using them anywhere else but, considering the initial points will likely never change or be used anywhere else, we're fine to just hard code 0L on line 65. (this also applies to initial score and initial time.

src/main/java/com/google/plantasy/utils/UserUtils.java
```diff
@@ -138,9 +140,9 @@ public static boolean addGameToUser(Entity userEntity, DatastoreService datastor
         log.severe("found empty gameId trying to add game to user " + (String) userEntity.getProperty("userID"));
         return false;
     }
- 
-    userEntity.setProperty("gameId", gameId);
-    datastore.put(userEntity);
+
+    User user = new User(userEntity);
+    user.setGame(gameId);
```
devonwasson, line 145: don't we still want to do a datastore.put()?

src/main/java/com/google/plantasy/utils/UserUtils.java
```diff
@@ -163,20 +165,22 @@ public static boolean addBlobKey(String blobKey, Entity userEntity, DatastoreSer
         return false;
     }
     
-    userEntity.setProperty("blobKey", blobKey);
-    datastore.put(userEntity);
- 
+    User user = new User(userEntity);
+    user.setBlobKey(blobKey);
```
devonwasson, line 169: don't we still want to do a datastore.put()?

src/main/java/com/google/plantasy/utils/UserUtils.java
```diff
@@ -163,20 +165,22 @@ public static boolean addBlobKey(String blobKey, Entity userEntity, DatastoreSer
...
+    User user = new User(userEntity);
+
     try {
-      userEntity.setProperty("score", ((Number) userEntity.getProperty("score")).intValue() + numPoints);
+      user.setScore(user.getScore() + numPoints);
     } catch (NullPointerException e) {
-      userEntity.setProperty("score", numPoints);                
+      user.setScore(numPoints);                
     }
     datastore.put(userEntity);
```
devonwasson, line 185: I think this needs to be changed from     datastore.put(userEntity);	to     datastore.put(user.GetUserEntity());	

Might want to check anywhere else we do datastore.put and make sure we're using the wrapper class and not the old entity.

src/main/java/com/google/plantasy/utils/UserUtils.java
```diff
@@ -205,11 +209,10 @@ public static void addPoints(Entity userEntity, int numPoints, DatastoreService
...
 
-    QuizTimingPropertiesUtils utils = new QuizTimingPropertiesUtils();
-    if(userEntity.getProperty("lastAwardedUploadPoints") == null || utils.isTimestampOutdated((long) userEntity.getProperty("lastAwardedUploadPoints"))){
+    User user = new User(userEntity);
+    if(user.getLastAwardedUploadPoints() == 0L || QuizTimingPropertiesUtils.isTimestampOutdated(user.getLastAwardedUploadPoints())){
         addPoints(userEntity, ADDED_POINTS, datastore);
-        userEntity.setProperty("lastAwardedUploadPoints", System.currentTimeMillis());
-        datastore.put(userEntity);
+        user.setLastAwardedUploadPoints(System.currentTimeMillis());
     }
```
devonwasson, line 216: don't we still want to do a datastore.put()?

src/main/java/com/google/plantasy/utils/UserUtils.java
```diff
@@ -163,20 +165,22 @@ public static boolean addBlobKey(String blobKey, Entity userEntity, DatastoreSer
...
+    User user = new User(userEntity);
+    user.setBlobKey(blobKey);
+
     return true; 
   }
  
   /**
     adds a specified number of points to the user's points
   */
   public static void addPoints(Entity userEntity, int numPoints, DatastoreService datastore){
```
devonwasson, line 177: AddPoints() should be changed to take in a User and not a UserEntity (this means modifying line 214 in this file to pass user and not userEntity.)

src/main/java/com/google/plantasy/utils/Game.java
EmmaWenzel: Got it, those are gone.

src/main/java/com/google/plantasy/utils/GameUtils.java
EmmaWenzel: That's done in all the places where an entity is updated.

src/main/java/com/google/plantasy/utils/GameUtils.java
EmmaWenzel: You're right, I took that out.

src/main/java/com/google/plantasy/utils/UserUtils.java
EmmaWenzel: Okay, combined those lines.

src/main/java/com/google/plantasy/utils/UserUtils.java
EmmaWenzel: Got it, that's done in all the places with datastore.put()

src/main/java/com/google/plantasy/utils/UserUtils.java
EmmaWenzel: Okay, that's done and changed a few places where it's called to match that

src/main/java/com/google/plantasy/utils/GameUtils.java
EmmaWenzel: Okay, that's done too + some changed places where that is called.

src/main/java/com/google/plantasy/utils/GameUtils.java
```diff
@@ -31,7 +31,7 @@
   /**
   * Creates a game entity and returns the entity if successful
   */
-  public static Entity createGameEntity(String gameName, DatastoreService datastore) {
+  public static Game createGameEntity(String gameName, DatastoreService datastore) {
```
devonwasson, line 34: This should be renamed to just createGame() considering we're no longer returning an entity. (I'm not sure if any other methods are now similarly misnamed but if any are, they should be fixed as well)

src/main/java/com/google/plantasy/utils/GameUtils.java
```diff
@@ -54,20 +54,13 @@ public static Entity createGameEntity(String gameName, DatastoreService datastor
     game.setQuizQuestion(quizQuestion);
     game.setQuizTimestamp(quiz_timestamp);
 
-    datastore.put(gameEntity);
+    datastore.put(game.getGameEntity());
```
devonwasson, line 57: We can remove this put as it's the same as the put on line 61

EmmaWenzel: Without this put I get an incomplete key error; I don't think the entity key is created until it is placed in datastore.

devonwasson: okay okay I see. 

So one quick thing, maybe move line 50 to be closer to line 59 because that's the first place we use that value. 

The other thing is... do we even need the gameId field? If the gameId == gameEntity.getKey then we can always get the game id by getting the key. That way, we don't have a redundant field in the datastore. 

We can still have a method in our Game wrapper class called getGameId() but it'll just return the keyToString(entity.getKey())

EmmaWenzel: Sounds good, this is out of scope for this pr but I'll follow up with another pr editing getGameId()

src/main/java/com/google/plantasy/utils/GameUtils.java
```diff
@@ -63,16 +62,6 @@ public static boolean addUserToGame(String userId, Entity gameEntity, DatastoreS
...
-    if(gameEntity == null){
-        log.severe("found null game entity trying to user "+ userId +" to game");
-        return false;
-    }
-    if(datastore == null){
-        log.severe("found null datastore trying to game to user " + userId);
-        return false;
-    }
-
-    Game game = new Game(gameEntity);
```
devonwasson: Does this run without this line (Game game = ...) ? We call game on line 69 but I don't think we define game in this function otherwise.

EmmaWenzel: I think I reverted some changes I shouldn't have thanks for catching that it all compiles and passes now.

PR #77, opened 2020-07-30
Fix null handling in tests 

colemanliyah:
changed the way I handled nulls in test cases; I got rid of try catch blocks in helper functions

### DIFF
--- a/src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
+++ b/src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.google.sps;
+package com.google.plantasy.utils;
 
 import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.DatastoreService;
@@ -34,7 +34,7 @@ public final class QuizTimingPropertiesUtils {
 
     private static final Logger log = Logger.getLogger(QuizTimingPropertiesUtils.class.getName());
 
-    static final List<String> quiz_questions = new ArrayList<String>(List.of(
+    static final List<String> quiz_questions = Arrays.asList(
         "Which plant has the most food growing from it?",
         "Which plant has the prettiest colors?",
         "Which plant is likely to grow the fastest?",
@@ -53,7 +53,7 @@ public final class QuizTimingPropertiesUtils {
         "Which plant would look the best inside as a houseplant?",
         "Which plant would look the best outside in a garden?",
         "Which plant would you give as a gift?"
-    ));
+    );
 
     //This function gets the the "quiz_timestamp" property of the entity that is fed into the function
     public static Long getQuizTimestampProperty(String entity, String id_of_entity, String id_of_entity_value, DatastoreService datastore) {
--- a/src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
+++ b/src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
  
-package com.google.plantasy.utils;
+package com.google.sps;
  
 import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.DatastoreService;
@@ -27,14 +27,14 @@ import java.util.logging.Logger;
 import java.util.logging.Level;
 import java.util.TimeZone;
 import java.text.SimpleDateFormat;
-import java.util.Arrays;
-
+import com.google.plantasy.utils.UserUtils;
+import com.google.plantasy.utils.Game;
+ 
 public final class QuizTimingPropertiesUtils {
-
+ 
     private static final Logger log = Logger.getLogger(QuizTimingPropertiesUtils.class.getName());
-
-    static final List<String> quiz_questions = Arrays.asList(
-        "Which plant has the most food growing from it?",
+ 
+    static final List<String> quiz_questions = new ArrayList<String>(List.of(
         "Which plant has the most food growing from it?",
         "Which plant has the prettiest colors?",
         "Which plant is likely to grow the fastest?",
@@ -52,17 +52,14 @@ public final class QuizTimingPropertiesUtils {
         "Which plant will most likely impress your friends and family?",
         "Which plant would look the best inside as a houseplant?",
         "Which plant would look the best outside in a garden?",
-        "Which plant would you give as a gift?");
+        "Which plant would you give as a gift?"
+    ));
+ 
     //This function gets the the "quiz_timestamp" property of the entity that is fed into the function
     public static Long getQuizTimestampProperty(String entity, String id_of_entity, String id_of_entity_value, DatastoreService datastore) {
         Query query = new Query(entity);
-        PreparedQuery pq;
-        try {
-            pq = datastore.prepare(query);
-        } catch(NullPointerException e) {
-            log.log(Level.SEVERE, "Null result for parameter {0}", datastore);
-            return null;
-        }
+        PreparedQuery pq = datastore.prepare(query);
+ 
         for(Entity query_entity : pq.asIterable()){
             if(query_entity.getProperty(id_of_entity).equals(id_of_entity_value)) {
                 return (Long) query_entity.getProperty("quiz_timestamp");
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
```diff
@@ -11,9 +11,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
- 
-package com.google.plantasy.utils;
```
EmmaWenzel, line 15: oh wait can you change this back to plantasy.utils

src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
```diff
@@ -27,14 +27,14 @@
...
 import java.text.SimpleDateFormat;
-import java.util.Arrays;
+import com.google.plantasy.utils.UserUtils;
+import com.google.plantasy.utils.Game;
 
 public final class QuizTimingPropertiesUtils {
 
     private static final Logger log = Logger.getLogger(QuizTimingPropertiesUtils.class.getName());
 
-    static final List<String> quiz_questions = Arrays.asList(
```
EmmaWenzel, line 36: also change this function back to this (I think it should work as long as you also get rid of the second parenthesis at the bottom)

src/test/java/com/google/plantasy/TimingPropertiesTest.java
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package com.google.plantasy;
```
EmmaWenzel, line 15: again here -- change it to plantasy (these might be merge conflicts?)

src/test/java/com/google/plantasy/TimingPropertiesTest.java
```diff
@@ -25,7 +25,7 @@
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
-import com.google.plantasy.utils.QuizTimingPropertiesUtils;
```
EmmaWenzel, line 28: ^ same as above

src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
colemanliyah: Sorry about this, I put all of these back!

src/test/java/com/google/plantasy/TimingPropertiesTest.java
```diff
@@ -50,61 +50,76 @@ public void getQuizTimestampProperty_nullUserTimeStamp() {
...
- 
-        Long actual = timing_properties_test.getQuizTimestampProperty("user", "userID", (user.getProperty("userID")).toString(), null);
-        Assert.assertEquals(null, actual);
+
+        Long actual = null;
+        try {
+            actual = timing_properties_test.getQuizTimestampProperty("user", "userID", (user.getProperty("userID")).toString(), null);
+        } catch(NullPointerException e) {
+            Assert.assertEquals(null, actual);
+        }
```
ofk-goog, line 92: Seems like what you want is to assert that a NullPointerException gets thrown, right? If so, try this: https://stackoverflow.com/questions/156503

Currently this test doesn't assert anything. (Neither do the other similar tests.) The only way for it to fail is if `getQuizTimestampProperty` throws an exception that isn't a `NullPointerException`.

src/test/java/com/google/plantasy/TimingPropertiesTest.java
```diff
@@ -50,61 +50,76 @@ public void getQuizTimestampProperty_nullUserTimeStamp() {
...
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
         datastore.put(user);
- 
-        Long actual = timing_properties_test.getQuizTimestampProperty("user", "userID", (user.getProperty("userID")).toString(), datastore); 
-        Assert.assertEquals(null, actual);
+
+        Long actual = null;
+        try {
+            actual = timing_properties_test.getQuizTimestampProperty("user", "userID", (user.getProperty("userID")).toString(), datastore); 
+        } catch (NullPointerException e) {
```
ofk-goog, line 60: I don't think you're testing what you intend to test here. When the `quiz_timestamp` in datastore is `null`, it makes sense for `getQuizTimestampProperty` to return `null`, right? Because `null` is the correct answer to "What is timestamp for this quiz?"

So there's no need for the try-catch block here. If this input causes a `NullPointerException`, the test ought to fail.

src/test/java/com/google/plantasy/TimingPropertiesTest.java
```diff
@@ -50,61 +50,76 @@ public void getQuizTimestampProperty_nullUserTimeStamp() {
...
+            Assert.assertEquals(null, actual);
+        }
```
devonwasson, line 62: here (and everywhere else in this PR) we'll want to actually assert that the exception was thrown. 

What you did is technically a correct way of accomplishing this goal, but it's not intuitive and there is a more elegant way to do this:

https://howtodoinjava.com/junit5/expected-exception-example/

Let me know if you have any questions!

src/test/java/com/google/plantasy/TimingPropertiesTest.java
devonwasson: omg I'm sorry I thought I was on this PR and not you, Oliver. Good news is we gave the exact same advice lol.

src/test/java/com/google/plantasy/TimingPropertiesTest.java
colemanliyah: *insert thumbs up emoji*

src/test/java/com/google/plantasy/TimingPropertiesTest.java
colemanliyah: insert thumbs up emoji

src/test/java/com/google/plantasy/TimingPropertiesTest.java
colemanliyah: Yes I took this out! This is why some tests will still assert nulls while the others have exception checks

src/test/java/com/google/plantasy/TimingPropertiesTest.java
```diff
@@ -275,21 +234,16 @@ public void giveUserQuizTakenPoints_notScoreProperty() {
...
 
-        boolean actual = false;
-        try{
-            actual = timing_properties_test.giveUserQuizTakenPoints(true, user, datastore);
-        } catch(NullPointerException e) {
-            user.setProperty("score", 20);
-        }
-        Assert.assertEquals(user.getProperty("score"), 20);
+        boolean actual = timing_properties_test.giveUserQuizTakenPoints(true, user, datastore);
+        user.setProperty("score", 20);
```
ofk-goog, line 238: This line should be removed, I think

src/test/java/com/google/plantasy/TimingPropertiesTest.java
```diff
@@ -275,21 +234,16 @@ public void giveUserQuizTakenPoints_notScoreProperty() {
...
     }
 
     @Test
     //Test if user was given points if their initial value was not zero
     public void giveUserQuizTakenPoints_actuallyAddedPoints_notStartAtZero(){
         QuizTimingPropertiesUtils timing_properties_test = new QuizTimingPropertiesUtils();
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-
+
```
ofk-goog: More pesky trailing whitespace here

colemanliyah: Yeah I got rid of these in the next commit

src/test/java/com/google/plantasy/TimingPropertiesTest.java
colemanliyah: Like 238 is there to add the score entity because it was not previously there

src/test/java/com/google/plantasy/TimingPropertiesTest.java
```diff
@@ -225,38 +219,38 @@ public void giveUserQuizTakenPoints_falseUserQuizStatus(){
...
-        
+
         Entity user = new Entity("user");
         user.setProperty("quiz_timestamp", 159430944365L);
         datastore.put(user);
- 
+
         boolean actual = timing_properties_test.giveUserQuizTakenPoints(true, user, datastore);
-        Assert.assertEquals(user.getProperty("score"), 20);
+        user.setProperty("score", 20);
```
devonwasson, line 238: I'm a bit confused by this test case. We're asserting that there will be a null pointer exception, right? Will that most likely happen on line 237? If so, do we need line 238?

colemanliyah: Oliver mentioned this too, I will get rid of line 238. I added it because this test means the user does not have a score entity so I was going to add it here

src/test/java/com/google/plantasy/TimingPropertiesTest.java
colemanliyah: I will get rid of this line

PR #78, opened 2020-07-31
Sort leaderboard

EmmaWenzel:


src/main/webapp/gameBoard.html
```diff
@@ -54,7 +54,7 @@ <h5 id="users-name"> User Name </h5>
                     <p> User's Plant</p> 
                 </div>
                 <div class="column right">
-                    <h5 id="rank"> Your in Nth place! </h5>
+                    <h5 id="rank"></h5>
```
colemanliyah, line 57: You can probably get rid of this tag altogether now. I don't think we are using it for anything

src/main/java/com/google/plantasy/utils/UserUtils.java
```diff
@@ -32,13 +32,21 @@
...
     public static final int ADDED_POINTS = 20;
     private static final Logger log = Logger.getLogger(UserUtils.class.getName());
  
-   
+    static final Comparator<User> RANK = new Comparator<User>() {
+        @Override
+        public int compare(User a, User b) {
+            return Long.compare(b.getScore(), a.getScore());
+        }
+    };
```
colemanliyah, line 48: Do you have to write this function? Doesn't java already come with a library where you can compare objects?

colemanliyah: I would add some documentation for this

src/main/webapp/gameBoard.html
EmmaWenzel: Got it, I took that out.

src/main/java/com/google/plantasy/utils/UserUtils.java
EmmaWenzel: I added some documentation for this function. I think I'm going to need a function to do this because I am sorting by a specific attribute of the object, in a certain order.

src/main/webapp/gameBoard.html
```diff
@@ -54,7 +54,6 @@ <h5 id="users-name"> User Name </h5>
                     <p> User's Plant</p> 
                 </div>
                 <div class="column right">
-                    <h5 id="rank"> Your in Nth place! </h5>
                 </div>
```
ofk-goog, line 57: Is this "column right" div still used for anything?

PR #80, opened 2020-08-03
Entire Logout Functionality

hannahsalameh:
Invalidates the user's session, deletes the session ID from the user entity, and removes the cookie that's tracking log in

### DIFF
--- a/src/main/java/com/google/plantasy/servlets/LogoutServlet.java
+++ b/src/main/java/com/google/plantasy/servlets/LogoutServlet.java
@@ -10,6 +10,7 @@ import javax.servlet.http.HttpServletResponse;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
+import com.google.plantasy.utils.User;
 import com.google.plantasy.utils.UserUtils;
 import com.google.plantasy.utils.CookieUtils;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/plantasy/servlets/LogoutServlet.java
```diff
@@ -0,0 +1,35 @@
+package com.google.sps.servlets;
+
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
```
devonwasson, line 5: please don't import *, only import what you need.

src/main/java/com/google/plantasy/servlets/LogoutServlet.java
```diff
@@ -0,0 +1,35 @@
+package com.google.sps.servlets;
```
devonwasson, line 1: wrong package name

src/main/java/com/google/plantasy/servlets/LogoutServlet.java
```diff
@@ -0,0 +1,35 @@
...
+import com.google.plantasy.utils.CookieUtils;
+
+@WebServlet("/logout")
+public class LogoutServlet extends HttpServlet {
+DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+      Cookie[] cookies = request.getCookies();
+      Entity userEntity = UserUtils.getUserFromCookie(cookies, datastore);
```
devonwasson, line 22: Should this be an entity or a User class?

src/main/java/com/google/plantasy/servlets/LogoutServlet.java
```diff
@@ -0,0 +1,35 @@
...
+      datastore.put(userEntity);
+
+      HttpSession session = request.getSession(false);
+      if(session != null){
+          session.invalidate();
+      }
+
+      Cookie cookie = CookieUtils.getCookieFromName(cookies,"SessionID");
+      cookie.setMaxAge(0); 
+      response.addCookie(cookie);
```
devonwasson, line 33: could you add a line documenting what this does?

src/main/java/com/google/plantasy/servlets/LogoutServlet.java
hannahsalameh: the user class doesnt handle session id's rn so it has to b an entity

devonwasson: gotcha, I would maybe add that to the user Entity in this CL so we can make that change. We shouldnt be introducing new references to the Entity class.

hannahsalameh: ok cool ill edit it

PR #81, opened 2020-08-04
Fix game

EmmaWenzel:
Change the game id to be gotten by the getGame function instead of being stored in the entity as a property

### DIFF
--- a/src/test/java/com/google/plantasy/GameTest.java
+++ b/src/test/java/com/google/plantasy/GameTest.java
@@ -76,7 +76,7 @@ public final class GameTest {
         try{
             datastore_game_entity = datastore.get(KeyFactory.stringToKey(game.getGameId()));
         }catch(Exception e){
-            Assert.assertEquals(game.getGameId(), KeyFactory.keyToString(datastore_game_entity.getKey()));
+            Assert.fail("Entity not found error thrown unexpectedly");
         }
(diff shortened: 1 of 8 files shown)


src/test/java/com/google/plantasy/GameTest.java
```diff
@@ -69,12 +68,16 @@ public void SetsGameEntityCorrectly() {
...
 
-        Entity datastore_game_entity = UserUtils.getEntityFromDatastore("Game", "gameId", game.getGameId(), datastore);
-        Assert.assertEquals(game.getGameId(), datastore_game_entity.getProperty("gameId"));
+        Entity datastore_game_entity = null;
+        try{
+            datastore_game_entity = datastore.get(KeyFactory.stringToKey(game.getGameId()));
+        }catch(Exception e){
+            Assert.assertEquals(game.getGameId(), KeyFactory.keyToString(datastore_game_entity.getKey()));
+        }
+        Assert.assertEquals(game.getGameId(), KeyFactory.keyToString(datastore_game_entity.getKey()));
```
devonwasson, line 81: I'm a bit confused by these two asserts. It looks like they do the same thing. Do we need both of them? Is one more important than the other? Do we expect an exception to be thrown? If we expect an exception, we should test for that explicitly.

EmmaWenzel: Switched the first Assert to fail

PR #84, opened 2020-08-04
Added login + logout functionality between pages

hannahsalameh:
(Basically splits loginScript.js into 2) Adds calls to the login and logout functions as well. 

Also, moved web.xml to its correct location as well as the images folder.


src/main/webapp/loggedin/logout.js
```diff
@@ -0,0 +1,30 @@
...
+document.addEventListener('DOMContentLoaded', (event) => {
+    var usernameElem = document.getElementById("username");
+    if(usernameElem){
+        usernameElem.innerHTML = localStorage.getItem("username");
+    }
+});
+ 
+//signs out of the page
+function signOut() {
+    //log out page --> DOESNT RUN FOR SOME REASON??
```
devonwasson, line 11: is this comment still relevant?

src/main/webapp/loggedin/logout.js
```diff
@@ -0,0 +1,30 @@
...
+ 
+    //log out google
+  gapi.load('auth2', function() {
+       gapi.auth2.init({
+         client_id: '610251294652-9ojdhjhh8kpcdvdbmrvp9nkevgr2n2d8.apps.googleusercontent.com'
```
devonwasson, line 16: I don't think we want this key to be stored here...

src/main/webapp/loginScript.js
```diff
@@ -1,50 +1,22 @@
...
-    }).then(function(auth2){
-        auth2.signOut();
-        localStorage.removeItem("username");
-        window.location = "login.html";
-    });
-  });
-}
-
-//current method to check sign in
-function checkSignIn(){
```
devonwasson, line 40: don't we still nee the checkSignIn() function?

hannahsalameh: no, this has been replaced with the filter and session ID method of checking pages

src/main/webapp/loggedin/logout.js
hannahsalameh: oh shoot yea ur right

src/main/webapp/loggedin/logout.js
```diff
@@ -8,12 +8,12 @@ document.addEventListener('DOMContentLoaded', (event) => {
...
 //signs out of the page
 function signOut() {
-    //log out page --> DOESNT RUN FOR SOME REASON??
+    var api_token = getToken();
  
     //log out google
   gapi.load('auth2', function() {
        gapi.auth2.init({
-         client_id: '610251294652-9ojdhjhh8kpcdvdbmrvp9nkevgr2n2d8.apps.googleusercontent.com'
+         client_id: api_token
```
devonwasson: I would just combine lines 11 and 16.

PR #85, opened 2020-08-04
Fixing the timing element

colemanliyah:
The user's time is now compared to today's time. Also the code was fixed to properly give points to the person who took the quiz

### DIFF
--- a/src/main/java/com/google/plantasy/servlets/QuizPoints.java
+++ b/src/main/java/com/google/plantasy/servlets/QuizPoints.java
@@ -16,14 +16,14 @@ import com.google.plantasy.utils.User;
 import java.io.*;
 import javax.servlet.*;
 import javax.servlet.http.*;
- 
+
 //This servlet's request comes from the quiz.html page; the request is the ID of the user who was voted for in a quiz
 //This servelt's functionality adds points the user who took a quiz and adds points the person who is voted for in a quiz
 @WebServlet("/quiz-points")
 public class QuizPoints extends HttpServlet {
- 
+
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
- 
+
     @Override
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
  
@@ -50,7 +50,7 @@ public class QuizPoints extends HttpServlet {
         Entity user_clicked = UserUtils.getEntityFromDatastore("user", "userID", clicked_user_id, datastore);
         User user_click = new User(user_clicked); 
         UserUtils.addPoints(user_click, 20, datastore);
- 
+
        response.sendRedirect("/loggedin/gameBoard");
     }
 }
--- a/src/main/java/com/google/plantasy/servlets/QuizPoints.java
+++ b/src/main/java/com/google/plantasy/servlets/QuizPoints.java
@@ -44,7 +44,6 @@ public class QuizPoints extends HttpServlet {
         current_user.setQuizTiming(System.currentTimeMillis());
         datastore.put(current_user.getEntity());
         
- 
         //Adds points to the user who got voted for in the quiz 
         String clicked_user_id = HttpRequestUtils.getParameterWithDefault(request, "user_picture", "");
         Entity user_clicked = UserUtils.getEntityFromDatastore("user", "userID", clicked_user_id, datastore);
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/plantasy/servlets/QuizPoints.java
```diff
@@ -37,17 +37,18 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+        current_user.setQuizTiming(System.currentTimeMillis());
+        datastore.put(current_user.getEntity());
+        
         //Adds points to the user who got voted for in the quiz 
         String clicked_user_id = HttpRequestUtils.getParameterWithDefault(request, "user_picture", "");
         Entity user_clicked = UserUtils.getEntityFromDatastore("user", "userID", clicked_user_id, datastore);
-        User user = new User(user_clicked);
-        UserUtils.addPoints(user, 20, datastore);
+        User user_click = new User(user_clicked); 
+        UserUtils.addPoints(user_click, 20, datastore);
```
devonwasson, line 51: nit: should be user_clicked or selected_user or something like that.

src/main/java/com/google/plantasy/servlets/UserQuizStatusServlet.java
```diff
@@ -1,32 +1,29 @@
...
-import java.util.Date;
-import com.google.appengine.api.datastore.Key;
-import com.google.plantasy.QuizTimingPropertiesUtils;
-import com.google.plantasy.UserUtils;
- 
+import com.google.plantasy.utils.QuizTimingPropertiesUtils;
+import com.google.plantasy.utils.UserUtils;
+import java.io.*;
+import javax.servlet.*;
+import javax.servlet.http.*;
```
devonwasson, line 16: please import specifically what you need rather than *

src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
```diff
@@ -71,14 +71,22 @@ public static Long getQuizTimestampProperty(String entity, String id_of_entity,
...
         sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
-        String users_quiz_time = sdf.format(usersQuizTime);
-        String current_quiz_time = sdf.format(currentQuizTime);
+        String users_quiz_time;
+        String today_time;
 
-        return(users_quiz_time.compareTo(current_quiz_time) > 0);
+        try {
+            users_quiz_time = sdf.format(usersQuizTime);
+            today_time = sdf.format(new Date());
```
devonwasson, line 83: we want to make today_time outside of the try catch block. There is no reason for this to fail so we don't need to try catch it.

src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
```diff
@@ -71,14 +71,22 @@ public static Long getQuizTimestampProperty(String entity, String id_of_entity,
...
+        } catch(IllegalArgumentException e ) {
+            log.log(Level.SEVERE, "Null result for parameter");
+            return false;
+        }  
+
+        return (users_quiz_time.equals(today_time));
```
devonwasson, line 89: doing equals works here but the logic is a bit confusing. if i understand correctly, this is allowed because the user quiz time will never be greater than today's date (because the user can't time travel) so just checking to see if it is equal is sufficient. If this is the case, maybe add a small comment stating as such to avoid confusion. 

Also, you can ditch the extra pair on () on this line (line 89)

src/main/java/com/google/plantasy/servlets/QuizPoints.java
colemanliyah: Changed this!

src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
colemanliyah: Got it!

src/main/java/com/google/plantasy/utils/QuizTimingPropertiesUtils.java
colemanliyah: Added some comments, let me know if that makes it more clear.

src/main/java/com/google/plantasy/servlets/UserQuizStatusServlet.java
colemanliyah: Fixed this!

PR #86, opened 2020-08-04
Fix timestamp naming

EmmaWenzel:


src/main/java/com/google/plantasy/utils/User.java
```diff
@@ -65,11 +65,11 @@ public void setScore(int score){
...
   // time the last quiz was taken
-  public long getQuizTiming(){
-    return (long) entity.getProperty("quiz_timing");
+  public long getQuizTimestamp(){
+    return (long) entity.getProperty("quiz_timestamp");
   }
-  public void setQuizTiming(long quiz_timing){
-    entity.setProperty("quiz_timing", quiz_timing);
+  public void setQuizTimestamp(long quiz_timestamp){
+    entity.setProperty("quiz_timestamp", quiz_timestamp);
```
devonwasson, line 72: It's interesting to me that we have these getters and setters yet we never use them in the code. Do we need them? I would assume we do but they're currently not being used anywhere, AFAIK.

EmmaWenzel: Hannah and Liyah both use these in their code I think, currently they're still using setProperty but those will be refactored to use these functions (if there's time).

PR #88, opened 2020-08-05
Fixingtests

colemanliyah:
Add three test for the new userHasTakenQuiz() function. Also got rid of instantiating QuizTimingProperties since all the methods I call are static. Some functions may look different because they got reverted back to old changes so I added them back in here

### DIFF
--- a/src/test/java/com/google/plantasy/TimingPropertiesTest.java
+++ b/src/test/java/com/google/plantasy/TimingPropertiesTest.java
@@ -132,6 +132,7 @@ public final class TimingPropertiesTest {
         Entity user = new Entity("user");
         DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
+        //Don't Change! The current system time will be converted to today's date
         user.setProperty("quiz_timestamp", System.currentTimeMillis());
         user.setProperty("userID", "abc");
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/plantasy/TimingPropertiesTest.java
```diff
@@ -46,139 +46,133 @@ public void tearDown() {
...
 
-        game.setQuizTimestamp(1594309443660L);
-        datastore.put(game.getGameEntity());
+    @Test
+    //Tests valid string, timestamp parameters for true test case
+    public void userTookQuiz_validParametersTrue(){
+        Entity user = new Entity("user");
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+        user.setProperty("quiz_timestamp", System.currentTimeMillis());
```
devonwasson, line 135: Typically, when testing, we don't want to use dynamic values like this. However, as you discovered, we need to do that here. 

I would suggest leaving a comment stating not to change this line and the reason why. Otherwise, someone might come in here, say "hmm this should be some hard coded value like in the function above", change it, run the test, discover it works (because maybe they used today's timestamp), then tomorrow it breaks again.

PR #89, opened 2020-08-06
fix issue where users without images break the quiz

devonwasson:
I felt bad breaking the demo so I fixed it

For reference, lines 41 through 52 would have been a good candidate for pulling into a helper function. That way, we could have done proper testing and might have caught the error where a user has no blobkey associated with their entry in the datastore.

### DIFF
--- a/src/main/java/com/google/plantasy/servlets/getUserImagesForQuizPage.java
+++ b/src/main/java/com/google/plantasy/servlets/getUserImagesForQuizPage.java
@@ -43,11 +43,9 @@ public class getUserImagesForQuizPage extends HttpServlet {
         for(String player_name : (ArrayList<String>) current_game.getProperty("userIds")) {
             if(!userEntity.getProperty("userID").equals(player_name)) {
                 User player = new User(UserUtils.getEntityFromDatastore("user","userID", player_name, datastore));
-                try {
-                    user_ids_and_pictures.put(player.getName(), player.getBlobKey());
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/plantasy/servlets/getUserImagesForQuizPage.java
```diff
@@ -39,15 +40,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-            if(!userEntity.getProperty("userID").equals(player)) {
-                Entity playerEntity = UserUtils.getEntityFromDatastore("user","userID", player, datastore);
-                user_ids_and_pictures.put(player, (playerEntity.getProperty("blobKey")).toString());
+        for(String player_name : (ArrayList<String>) current_game.getProperty("userIds")) {
+            if(!userEntity.getProperty("userID").equals(player_name)) {
+                User player = new User(UserUtils.getEntityFromDatastore("user","userID", player_name, datastore));
+                try {
+                    user_ids_and_pictures.put(player.getName(), player.getBlobKey());
+                } catch (Exception e){
+                    // If player.getBlobKey() causes a NPE, we can skip that user.
```
ofk-goog, line 49: IIUC `player.getBlobKey()` will return `null` rather than throw a NPE. And adding a null value to a HashMap is also legal. So instead I think we would need to check for a null blob key in the `if` condition.

